PR #254, opened 2023-03-28
Disabled costly clang-tidy checks

mihe:
In an effort to bring down the time spent by the CI builds I'm disabling [`bugprone-reserved-identifier`](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-reserved-identifier.html) and [`misc-const-correctness`](https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/const-correctness.html).

While I don't really care much about stomping on reserved identifiers, the const-correctness one does sting a bit, but it single-handedly almost doubles the time spent by clang-tidy, which is by far the most costly workflow in the CI runs, so I don't have much choice but to get rid of it. I've also found it to leave quite a lot of variables untouched that could have been `const`, so I'm not sure how much good it's actually doing.


mihe:
I forgot that I had originally written this clang-tidy config for LLVM 15, and then downgraded it to LLVM 14 just enough to make it run, but while leaving the checks in from LLVM 15. `misc-const-correctness` is an LLVM 15 check and as such has never run as part of the CI checks, so CI runs barely saw any speed improvements from this change, sadly.

Either way, LLVM 15 is just around the corner for Xcode (14.3), which means I'll be bumping compiler versions pretty soon, so these checks need to go.

PR #321, opened 2023-05-08
Replaced clang-tidy with SonarCloud

mihe:
The time it takes for clang-tidy to analyze stuff has started to get to me, so I'm trying out SonarCloud as a replacement.

SonarCloud seems to find some stuff that clang-tidy doesn't pick up on, so I'm hoping this will be an improvement.


mihe:
Getting SonarCloud to run automatically as part of a pull request quickly became a massive headache, since they need the `SONAR_TOKEN` secret inside an inherently unsafe environment.

I can work around it by chaining workflows and other such things, but I don't think it's worth complicating the CI workflows to that degree just for this, so I'll leave this for now.

I can always run SonarCloud manually every now and then and resolve whatever issues that clang-tidy doesn't catch.

PR #398, opened 2023-05-25
Fixed issue with contact velocities lagging behind one tick

### LINKED ISSUE #388: Contact velocities lag behind by one tick
Hi, I've been sketching out a platformer controller with this physics engine, I've tried similar things with Godot's built in physics engine and bullet in Godot 3 and so far I think this is much better than bullet overall, 4.0's physics are not usable for this I think. 

First I was trying to replicate something I had tried in bullet and I noticed the moving platform was delayed by a frame, this works by applying a translation to the body's direct's state in _integrate_forces(), the velocity is obtained by searching the floor using contact monitoring and CI, this might not be the intended way to use the engine, but it's the best thing I could figure out so far. 

https://github.com/godot-jolt/godot-jolt/assets/18746022/afd1579c-8a91-41f0-ba14-d1240f58a181

Then I tried a rotating platform that tries to do the same thing, but fails with the cylinder shape, it works better with a capsule shape, probably because the collider's contact is near the bottom center of the shape, but I'm not sure how to get it working properly without hardcoding the logic, I'm not experienced with 3d physics enigines so I'm not sure if this is a real problem and I didn't try that in bullet so I don't have a point of reference.

The second set of moving platforms fix this in code, the linear platform fixes the 1 frame delay by calculating the velocity in _physics_process and character controller uses that value instead,  this didn't happen with bullet in Godot 3.5. The rotating one calculate's the player's velocity using the using the bottom center of the player as reference instead of the velocity for the contact point reported by the engine.

Here's the project:
[MJ.zip](https://github.com/godot-jolt/godot-jolt/files/11540134/MJ.zip)

mihe:
Fixes #388.

The contact velocities have until now been populated using `JPH::Body::GetPointVelocity` within the contact listener. This turned out to cause the velocity to lag behind one tick, which I'm guessing is because the constraint solving hasn't happened yet at that point, which in turn presumably affects the velocity.

So instead I decided to simply not have the velocities be part of the contacts and instead just get them from the collider as the user requests them.

### DIFF
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -351,7 +351,6 @@ void JoltBodyImpl3D::add_contact(
 	const Vector3& p_normal,
 	const Vector3& p_position,
 	const Vector3& p_collider_position,
-	const Vector3& p_collider_velocity,
 	const Vector3& p_impulse
 ) {
 	const int32_t max_contacts = get_max_contacts_reported();
@@ -386,7 +385,6 @@ void JoltBodyImpl3D::add_contact(
 		contact->normal = p_normal;
 		contact->position = p_position;
 		contact->collider_position = p_collider_position;
-		contact->collider_velocity = p_collider_velocity;
 		contact->impulse = p_impulse;
 	}
 }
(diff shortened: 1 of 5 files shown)


mihe:
I just realized this can't be right. You want the contact velocity to be at the time of contact, and not the current velocity. If a body hits a wall and comes to a dead stop, you want the velocity at which it hit the wall, not a zero vector.

Issue #379, opened 2023-05-22
Add body name to warning and error message


mgiraud:
There are a lot of usefull warning and errors that you have added to your extension.
However when a scene has a lot of physical objects, it's tedious to identify which object causes the warning/error.
If there is not too much performace cost, could you consider adding the name of the object to the message ?

Edit : Printing the hierarchy would be the best, but maybe a bad idea in regards to readability

mihe:
I can only do this for a very limited set of warnings/errors, due to the limited amount of information I have from the physics server. Specifically the ones related to bodies, like warnings about bounce, friction and linear/angular damping. I don't have any sort of instance ID or other reference for things like joints, so I have no way of knowing what object in Godot is using that particular thing on my side.

Hierarchy would technically be possible for those, I think, but I'm not sure what edge-cases there might be in terms of traversing the tree.

Are there any warnings/errors in particular that you run into a lot?

mgiraud:
I've had a lot of 
```
E 0:00:00:0754   build: Failed to build box shape with {half_extents=(0.150000, 0.150000, 0.350000) margin=5.957000}. Its half extents must be greater than its margin.
  <Erreur C++>   Condition "shortest_axis <= margin" is true. Returning: {}
  <Source C++>   shapes\jolt_box_shape_impl_3d.cpp:43 @ build()
```
or similar during my first build because Godot physics does not use this value ^^

But i have this error that i have trouble fixing, i can't find the faulty physics material  
```
W 0:00:00:0745   set_bounce: Bounce values less than 0 or greater than 1 are not supported by Godot Jolt. Values outside this range will be clamped.
  <Source C++>   objects\jolt_body_impl_3d.cpp:821 @ set_bounce()
```

mihe:
I could technically print the names of the bodies that own a particular shape. The problem with that is that you can reuse the same shape across multiple bodies, so you'd potentially have dozens/hundreds/thousands of bodies per message, which would not only be slow to print but also incredibly noisy, much like printing the node hierarchy.

It also wouldn't help with bespoke shapes like the ones you'd create for things like shape-casts and other physics queries.

I'll see about adding the names of the bodies in the messages I mentioned earlier, but that's about as much as I can do I think.

mgiraud:
That would be plenty enough, thanks !

mihe:
[...]

PR #406, opened 2023-05-28
Changed shape margins to be interpreted as an upper bound

mihe:
As discussed in #379.

This removes the warnings, for boxes and cylinders, that are emitted when the margin is greater than its minimum half extent.

The margin is now instead interpreted as an upper bound, and computed based on an arbitrary factor of the shape's minimum half extent.


phelioz:
Really nice, thanks! :relaxed:

mihe:
Tack själv!

It's always tricky to know what people will rub up against. I figured these warnings would be a fairly rare sight, but here we are, this close to the first release. :)

phelioz:
Must again also say thanks for the addon. Have had problems with performance and stability with the 3D physics in Godot and this addon is a real improvement.

PR #457, opened 2023-06-18
Added collision response override for asymmetrical collisions

### LINKED ISSUE #400: Collision layers/masks differ in behavior from Godot Physics
Hello,

Creating this issue to discuss the differences between Godot collision_layer / masks and those provided by Godot Jolt because of discussion in https://github.com/jrouwe/JoltPhysics/discussions/544.

I've looking through the documentation a bit and I'm wondering what exactly would be needed from Jolt in order to support this. As far as I understand Godot Jolt does:

colliding = (layerA & maskB) OR (layerB & maskA)

When rigid bodies collide they interact normally.

whereas Godot does:

colliding = layerA & maskB

for normal collision queries and when rigid bodies collide we have:

| layerA & maskB  | layerB & maskA | Result |
| ------------- | ------------- | ------------- |
| False | False | Objects don't collide |
| True | True | Objects collide and interact normally |
| True | False | Objects collide and object B is treated as having infinite mass |
| False | True | Objects collide and object A is treated as having infinite mass |

My guess is that if both bodies end up having infinite mass (because the other was already kinematic) that the collision is discarded as well.

Does this summarize things correctly?

As far as I can see there is no issue for collision queries. ```ObjectLayerFilter::ShouldCollide``` can be implemented to do the same thing as Godot 4 does as you can store the layer of object A in the ObjectLayerFilter and just and it with the layer that is being passed in. Is this correct? (I'm hoping this is already the case now)

I think the only problem is for the interaction between rigid bodies. Jolt uses ```ObjectLayerPairFilter::ShouldCollide``` to determine if two objects should collide and there you are forced to return (layerA & maskB) OR (layerB & maskA) because this function only determines if they collide not how. It sounds like the only thing you actually need is the ability to override the masses of the colliding bodies? I haven't fully investigated this but it looks like it would be possible to either override the motion types for the bodies in ContactSettings (you would only be able to go in the direction Dynamic -> Kinematic -> Static and not e.g. promote a Static object to Dynamic). Alternatively there could be an 'inverse mass scale' for body 1 & 2, setting it to 0 would make the body have infinite mass, setting it to 1/2 would make it twice as heavy etc. Inverse inertias would scale accordingly.

What do you think about this?

mihe:
Fixes #400.

This changes collision layers/masks to behave less like they did in Godot 3 and more like they do in Godot 4, where having an asymmetric layer/mask setup will result in the body whose mask does not contain the layer of the other body effectively having infinite inertia/mass in the context of that collision.

### DIFF
--- a/src/spaces/jolt_contact_listener_3d.cpp
+++ b/src/spaces/jolt_contact_listener_3d.cpp
@@ -46,6 +46,8 @@ void JoltContactListener3D::OnContactAdded(
 	add_debug_contacts(p_manifold);
 #endif // GDJ_CONFIG_EDITOR
 
+	override_collision_response(p_body1, p_body2, p_settings);
+
 	apply_surface_velocities(p_body1, p_body2, p_manifold, p_settings);
 
 	if (!is_listening_for(p_body1) && !is_listening_for(p_body2)) {
@@ -73,6 +75,8 @@ void JoltContactListener3D::OnContactPersisted(
 #ifdef GDJ_CONFIG_EDITOR
 	add_debug_contacts(p_manifold);
 #endif // GDJ_CONFIG_EDITOR
+
+	override_collision_response(p_body1, p_body2, p_settings);
 
 	apply_surface_velocities(p_body1, p_body2, p_manifold, p_settings);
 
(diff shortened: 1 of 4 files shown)


src/spaces/jolt_contact_listener_3d.cpp
```diff
@@ -333,6 +337,55 @@ void JoltContactListener3D::flush_area_exits() {
...
+
+	if (p_body1.IsKinematic() || (second_scans_first && !first_scans_second)) {
+		p_settings.mInvMassScale1 = 0.0f;
+		p_settings.mInvInertiaScale1 = 0.0f;
+	}
+
+	if (p_body2.IsKinematic() || (first_scans_second && !second_scans_first)) {
+		p_settings.mInvMassScale2 = 0.0f;
+		p_settings.mInvInertiaScale2 = 0.0f;
+	}
```
mihe, line 386: @jrouwe Needing to explicitly override the mass/inertia for kinematic bodies here struck me as a bit strange. Shouldn't they by default have that already?

If I leave those conditions out the collision ends up being somewhat strange, where the non-kinematic body (with infinite mass/inertia) will still collide with the kinematic body, but end up penetrating it ever so slightly before coming to a full stop.

mihe: Here's a video of the scene from #400 showing the behavior when those `IsKinematic()` checks are omitted:

https://github.com/godot-jolt/godot-jolt/assets/4884246/24e42a3d-01ce-4e5a-9abc-9997ee6d0621

jrouwe: Yes, I didn't test this properly. I knew that the value would be unused for a kinematic body, but it was actually used in 1 place :)

The fix is in https://github.com/jrouwe/JoltPhysics/tree/bugfix/kinematic_vs_overridden_mass

I still need to do a bit of testing on it

jrouwe: Fix is submitted

Issue #390, opened 2023-05-23
Project settings have no descriptions within the editor


mihe:
There's currently no ability from GDExtension to add in-editor documentation, and as such we're forced to keep the documentation for the project settings in [a separate document](https://github.com/godot-jolt/godot-jolt/blob/master/docs/settings.md) within the repo.

There's currently an open pull request for this (godotengine/godot#75415), with ongoing discussions, so hopefully this will be resolved in a not too distant future.

mihe:
While support for GDExtension-provided documentation has been added to Godot now, this unfortunately only applies to new classes provided by the extension, meaning there's still no support for adding documentation to project settings.

With no pull request on the horizon for this, and with this extension having now entered into maintenance mode as part of being merged into Godot in the form of an engine module, it's highly unlikely that this will ever be addressed in this extension. As such, I'll go ahead and close this issue.

PR #497, opened 2023-07-15
Added `JoltPinJoint3D` along with editor/gizmo plugin

### LINKED ISSUE #346: Add a substitute node for `PinJoint3D`
In order to better line up with Jolt's `JPH::PointConstraint` there should be something like a `JoltPinJoint3D` node that omits the bias, damping, impulse clamp and solver priority.

mihe:
Fixes #346.

This adds a substitute node for `PinJoint3D`, called `JoltPinJoint3D`, that matches the `PinJoint3D` interface, while adding a couple of things on top, which are as follows:

- The ability to enable/disable the joint, using its `enabled` property
- The ability to fetch its last applied impulse, using its `get_linear_impulse` method
- The ability to override the solver velocity/position iterations for that specific joint

The first and second points together allow for creating breakable joints.

Note that this also introduces an editor plugin, which is responsible for setting the editor icons for these new joint nodes, as well as registering a new gizmo plugin, which draws the editor gizmos for them.

### DIFF
--- a/src/joints/jolt_cone_twist_joint_impl_3d.cpp
+++ b/src/joints/jolt_cone_twist_joint_impl_3d.cpp
@@ -136,6 +136,8 @@ void JoltConeTwistJointImpl3D::rebuild(bool p_lock) {
 	);
 
 	space->add_joint(this);
+
+	_update_enabled();
 }
 
 JPH::Constraint* JoltConeTwistJointImpl3D::_build_swing_twist(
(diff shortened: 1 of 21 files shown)


mihe:
I'm curious what you think of naming this `get_total_lambda_position` method after your Jolt equivalents, @jrouwe.

You describe these lambda positions/rotations in the documentation as representing "the linear/angular impulse applied at the constraint in the last physics step to keep the constraint together", so I'm wondering if maybe it would be okay/accurate to just name these `get_linear_impulse`/`get_angular_impulse` instead? I feel like that might be more intuitive to laymen like myself.

I can always clarify in the ([eventual](https://github.com/godot-jolt/godot-jolt/issues/390)) in-editor documentation for these methods that they represent the Lagrange multiplier and whatnot.

jrouwe:
I agree that linear/angular impulse is a better name for it

PR #559, opened 2023-08-21
Fixed issue with `move_and_slide` getting stuck while sliding

### LINKED ISSUE #529: CharacterBody3D sticks, stutters, slows down when sliding against walls
godot 4.1.1
godot-jolt 0.5.0-rc1

In my 3d platformer game the character sticks to, stutters on and slows down against walls.

I have attached a project that I hope will be useful for testing this.

[joltWallSlideTest0.zip](https://github.com/godot-jolt/godot-jolt/files/12244228/joltWallSlideTest0.zip)

Video (sped-up to trade off length versus file size):

[joltwallslidetest0demo.webm](https://github.com/godot-jolt/godot-jolt/assets/64640811/70541bc0-827e-4f6f-ba97-3facc0f00592)

When you press run, a thousand-odd characterbody3d's with box collision will walk across a section of trimesh level collision at some amount of slope into a wall at some angle. You can also click one for a closer look (click again to return). The project will run slow but I don't think (?) this will affect results. If desired I could add a smaller, more focused test scene.

Generally, characters should slide along walls until they fall off the level (except those on green sections with a flat wall). You'll see under a lot of typical configurations that some will get stuck. If you move them by repeatedly tapping up, most (but usually not all) will make progress along their wall. The most success appears when block on walls is disabled, but normally you'll want it enabled.

note: haven't tested under godot physics yet, but I expect it might be similar (I think I've seen some allusions to similar issues being more to do with move_and_slide than the physics itself?)

mihe:
Fixes #529.

This introduces a minimum penetration depth for the contacts reported by `test_body_motion`, similar to how Godot Physics does it, which alleviates problems with `move_and_slide` where even after recovery you could end up still overlapping with the wall you're sliding against ever so slightly, which `floor_block_on_wall` would interpret as being trapped in a corner and would subsequently stop the character.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ Breaking changes are denoted with ⚠️.
 
 - Fixed issue where ray-casting a `ConcavePolygonShape3D` that had `backface_collision` enabled, you
   would sometimes end up with a flipped normal.
+- Fixed issue where a `CharacterBody3D` using `move_and_slide` could sometimes get stuck when
+  sliding along a wall.
 
 ## [0.6.0] - 2023-08-17
 
--- a/src/spaces/jolt_physics_direct_space_state_3d.cpp
+++ b/src/spaces/jolt_physics_direct_space_state_3d.cpp
@@ -512,7 +512,14 @@ bool JoltPhysicsDirectSpaceState3D::test_body_motion(
 	Vector3 scale;
 	Transform3D transform = Math::decomposed(p_transform, scale);
 
-	const Vector3 direction = p_motion.normalized();
+	const float distance_sq = p_motion.length_squared();
+	float distance = 0.0f;
+	Vector3 direction = {};
+
+	if (distance_sq > 0.0f) {
+		distance = Math::sqrt(distance_sq);
+		direction = p_motion / distance;
+	}
 
 	Vector3 recovery;
 	const bool recovered = _body_motion_recover(p_body, transform, direction, p_margin, recovery);
@@ -539,6 +546,7 @@ bool JoltPhysicsDirectSpaceState3D::test_body_motion(
 			p_body,
 			transform.translated(p_motion * unsafe_fraction),
 			direction,
+			distance,
 			p_margin,
 			p_max_collisions,
 			p_result
@@ -883,6 +891,7 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 	const JoltBodyImpl3D& p_body,
 	const Transform3D& p_transform,
 	const Vector3& p_direction,
+	float p_distance,
 	float p_margin,
 	int32_t p_max_collisions,
 	PhysicsServer3DExtensionMotionResult* p_result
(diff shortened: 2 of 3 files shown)


yosoyfreeman:
Hi! I'm writing my own implementation of an sliding algorithm and i think the problem i found was introduced here. 

While this in fact improves the problem of ghost collisions, i'm having the opposite effect, in which the body is unable to find one of the collisions in an acute angle scenario like this:

 
imagen

The capsule in this shape is moving towards the ramp, therefore colliding with the floor and the ceiling and the same time. This is correctly reported by Godot physics but will fluctuate in Godot Jolt, i believe because the resolution in this acute angle scenarios can produce a really small displacement which is now ignored.

mihe:
Yes, there's a number of regression that have come from this change, unfortunately. I've spent a bit of time in the past few weeks trying to undo it, but the amount of ghost collisions it reintroduces is unacceptably high.

I can mitigate some of that by removing the `stop_all_motion` stuff seen [here](https://github.com/godotengine/godot/blob/61282068f4d59cb48f35ad95391728c58d9008ab/scene/3d/physics/character_body_3d.cpp#L268-L286) in `move_and_slide`, which I frankly don't understand the purpose of, but there's still plenty of ghost collisions left even after that that I have yet to understand why/where/how they arise.

yosoyfreeman:
Thank you a lot, allow me to put some light on the issue:

The current way in which move_and_slide works have serious design problems that are not directly related with the way collision is resolved. 

The problem of acute corners is that you could need thousands of even near infinite iterations to resolve the full motion, because the body collides with one plane, barely moves, touches the other, projects velocity, looses almost nothing and the cycle repeats. This is the origin of the character being stuck in past versions and in implementations of this algorithm in general.

This problem was explored later on in a new paper [Improving the Numerical Robustness of Sphere
Swept Collision Detection](https://arxiv.org/pdf/1211.0059.pdf)

This second paper is accurate in one thing, colliding with two planes only allows movement along the crease vector between the two and 3 collisions at the same time is a full stop, because we are left without any axis of freedom.

However, this paper made a mistake that has propagated to other implementations, including ours: The axis of freedom ONLY decrease if the collisions are SIMULTANEOUS. By limiting the axis of freedom based on the previous collision we are wrongly assuming the last plane extends infinitely towards the new one, converting the sliding algorithm in a collision resolution algorithm more suitable for direct Input. 

Which leave us to the current implementation, which tries without success to implement the fixes of the second paper. In this case the error of assuming the plane extends infinitely is there, but even more problematic, this is done only if the two planes are a wall, leaving us with a collision resolution that will behave totally different from what you would expect and will change behavior based in totally arbitrary parameters.

The part of the code you linked is one of the more problematic ones. Is trying to apply the fix to the corner case, but as a consequence of a lack of understanding of the fix itself what is really happening is that the sliding algorithm is not such thing in the case that you hit two walls during the same move and slide call, as it will not only break the trajectory by assuming the last wall was infinite, but will still keep the stutter because the floor does not have such checks.

So i fear you could try to fix test_move for years without success. Is the whole move and slide implementation which will always cause such stops and stutters when it should not.

This is the problem i been tackling for weeks and i created a new solution based on the research of this time which fixed the design problems of the two original papers and our current implementation.

Hope this can help.

PR #588, opened 2023-09-17
Added drawing of triangle outlines to `JoltDebugGeometry3D`

jrouwe:
This makes it easier to see which edges are active and which ones are not.


mihe:
Thank you! This is one of those things I've been manually adding to the code every time I want to use it, for reasons I can't explain. I should have just added the property instead.

I took some creative liberties and threw in the height map triangle outlines as well, as well as renaming the property to just `draw_triangle_outlines`. The getter ends up being a little bit strange now, but whatever, this node isn't really part of the public API anymore anyway.

Issue #587, opened 2023-09-16
Sphere shapes collide with inactive edges


yosoyfreeman:
I open another issue because https://github.com/godot-jolt/godot-jolt/issues/508 is indeed fixed.

A similar problem happens when a sphere is used, but in this case, it will collide between perfectly coplanar triangles too. 

https://github.com/godot-jolt/godot-jolt/assets/64601424/d877373b-2e74-432d-9910-694e7d63f8b3

This causes the ball to be launched against the floor. This is how it affects our game:

https://github.com/godot-jolt/godot-jolt/assets/64601424/7e0971ef-8b50-4a64-af37-de206d9a7cbe

The following minimal reproduction project contains two scenes equally setup with a box and a sphere for you to be able to compare quickly between the two.

[BallStuckBug.zip](https://github.com/godot-jolt/godot-jolt/files/12640936/BallStuckBug.zip)

mihe:
Yeah, that does not look right. I'll dig into it.

jrouwe:
@yosoyfreeman Are these triangles from the same mesh? And are you sure this edge is not part of a T-edge:

![image](https://github.com/godot-jolt/godot-jolt/assets/1621693/fd4d738d-389f-4acc-869c-33e910cc3fab)

Jolt can't detect active edges across different meshes and it currently treats all T-edges as active since there is are multiple edges that connects to that edge and it is currently not smart enough check if all neighbors are coplanar (mainly because it detects neighboring triangles by shared vertices and not if vertices happen to fall on an edge).

yosoyfreeman:
Hi! The collision shape is a trimesh collision shape generated from a single mesh. I thought about this kind of scenarios being a variable so this example uses the original shape from https://github.com/godot-jolt/godot-jolt/issues/508. The mesh was scaled up to have a bigger top plane and the collision shape re generated accordingly. I thought using the same mesh would produce less variables. In our game, the collision shapes are also trimeshs. Hope this helps.

jrouwe:
The mesh is indeed fine, the edge is not an active edge:

![image](https://github.com/godot-jolt/godot-jolt/assets/1621693/3df6c477-054f-4063-abe4-e51379da29c6)

For reference, the bottom of the shape has an active edge which is colored green:

![image](https://github.com/godot-jolt/godot-jolt/assets/1621693/364129c8-c266-4d17-9d1d-519145aa2160)

[...]

PR #590, opened 2023-09-17
Added ability to dump snapshots of physics spaces

jrouwe:
This snapshot can be used to report issues in Jolt Physics, it allows to easily load a problematic scene in the Samples framework.

### DIFF
--- a/src/shapes/jolt_shape_impl_3d.cpp
+++ b/src/shapes/jolt_shape_impl_3d.cpp
@@ -190,7 +190,65 @@ JPH::ShapeRefC JoltShapeImpl3D::without_custom_shapes(const JPH::Shape* p_shape)
 		case JoltCustomShapeSubType::OVERRIDE_USER_DATA:
 		case JoltCustomShapeSubType::DOUBLE_SIDED: {
 			// Replace unsupported decorator shapes with the inner shape
-			return static_cast<const JPH::DecoratedShape*>(p_shape)->GetInnerShape();
+
+			const auto* shape = static_cast<const JPH::DecoratedShape*>(p_shape);
+			const auto* inner_shape = shape->GetInnerShape();
+
+			return without_custom_shapes(inner_shape);
+		}
+
+		case JPH::EShapeSubType::StaticCompound: {
+			const auto* shape = static_cast<const JPH::StaticCompoundShape*>(p_shape);
+
+			JPH::StaticCompoundShapeSettings settings;
+
+			for (const auto& sub_shape : shape->GetSubShapes()) {
+				settings.AddShape(
+					sub_shape.GetPositionCOM(),
+					sub_shape.GetRotation(),
+					without_custom_shapes(sub_shape.mShape)
+				);
+			}
+
+			const JPH::ShapeSettings::ShapeResult shape_result = settings.Create();
+
+			ERR_FAIL_COND_D_MSG(
+				shape_result.HasError(),
+				vformat(
+					"Failed to recreate static compound shape during filtering of custom shapes. "
+					"It returned the following error: '%s'.",
+					to_godot(shape_result.GetError())
+				)
+			);
+
+			return shape_result.Get();
+		}
+
+		case JPH::EShapeSubType::MutableCompound: {
+			const auto* shape = static_cast<const JPH::MutableCompoundShape*>(p_shape);
+
+			JPH::MutableCompoundShapeSettings settings;
+
+			for (const auto& sub_shape : shape->GetSubShapes()) {
+				settings.AddShape(
+					sub_shape.GetPositionCOM(),
+					sub_shape.GetRotation(),
+					without_custom_shapes(sub_shape.mShape)
+				);
+			}
+
+			const JPH::ShapeSettings::ShapeResult shape_result = settings.Create();
+
+			ERR_FAIL_COND_D_MSG(
+				shape_result.HasError(),
+				vformat(
+					"Failed to recreate mutable compound shape during filtering of custom shapes. "
+					"It returned the following error: '%s'.",
+					to_godot(shape_result.GetError())
+				)
+			);
+
+			return shape_result.Get();
 		}
 
 		case JPH::EShapeSubType::RotatedTranslated: {
--- a/src/shapes/jolt_shape_impl_3d.cpp
+++ b/src/shapes/jolt_shape_impl_3d.cpp
@@ -178,6 +178,70 @@ JPH::ShapeRefC JoltShapeImpl3D::with_user_data(const JPH::Shape* p_shape, uint64
 	return shape_result.Get();
 }
 
+JPH::ShapeRefC JoltShapeImpl3D::without_custom_shapes(const JPH::Shape* p_shape) {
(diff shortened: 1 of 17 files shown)


jrouwe:
Hello,

I'm not sure if this is the best way of implementing things, but I created an option to make it easier to go from a repro in godot to something I can easily debug in the Samples framework. With this I'm able to load the scene of #587 to start investigating it.

It adds an extra option to the JoltDebugGeometry3D:

image

Which, when clicked produces a snapshot.bin in the current working folder.

I don't know if it is possible to add a button instead of a checkbox that turns off immediately again and I tried to follow your style as closely as possible, but I can imagine you don't want to incorporate it in this form. If you want some changes then let me know (or make them yourself 😄)!

jrouwe:
I see there's some failing stuff. I'll fix that in a couple of hours.

mihe:
That does look pretty useful!

There's unfortunately no way to add a button to the property inspector, as far as I'm aware. I've seen others use the same workaround in their custom nodes, but I think piggybacking on `JoltDebugGeometry3D` for this is pushing things a little bit.

The more appropriate option for this would probably be to make another editor plugin specifically for debug actions like this and add a menu button to something like the `Debug` menu or perhaps `Project > Tools`.

I can whip that up for you. Don't worry about the CI failures.

jrouwe:
Will it then still be possible to trigger it from GDScript? (In case the script needs to do something to complete the repro case)

src/spaces/jolt_debug_geometry_3d.cpp
```diff
@@ -115,8 +123,44 @@ JoltDebugGeometry3D::~JoltDebugGeometry3D() = default;
...
+			return p_shape;
+		}
+
+		case JPH::EShapeSubType::StaticCompound:
+		case JPH::EShapeSubType::MutableCompound:
+		case JPH::EShapeSubType::RotatedTranslated:
+		case JPH::EShapeSubType::OffsetCenterOfMass: {
+			ERR_PRINT("Unsupported shape type, snapshot may not be readable in Jolt!");
+			return p_shape;
+		}
```
mihe, line 156: What's with needing to special-case the decorator and compound shapes by the way? Is `SaveBinaryState` not able to deal with those types of shapes?

Both `RotatedTranslated` and `StaticCompound` will be quite prevalent. `OffsetCenterOfMass` not so much perhaps, but it's used as well.

mihe:
> Will it then still be possible to trigger it from GDScript? (In case the script needs to do something to complete the repro case)

Yes. I figured I would expose it as a method on the physics server and use that from the editor plugin. It won't be quite as slick as toggling a property on a `JoltDebugGeometry3D`, but it should work.

src/spaces/jolt_debug_geometry_3d.cpp
jrouwe: Yes SaveBinaryState would work but the Jolt Samples app doesn't know about to these shapes so will fail loading. As most shapes are immutable I need to create new shapes, and I only implemented the ones in the scene I had to export (not knowing if you would like to merge the code or not, if you do I can create code for the other cases too)

mihe: Oh right, you need to create new shapes because they could be wrapping/containing any of the custom shapes. That makes sense.

I'll see if I can implement them while I'm at it.

jrouwe:
Cool stuff! If you want I can do the compound shapes.

mihe:
So I just tried loading a snapshot of `examples/scenes/shapes/shapes.tscn` in `Samples`, and I end up hitting [this](https://github.com/jrouwe/JoltPhysics/blob/dd1d4a670c1dd73f7ec8dbaa0cabf9162a31e159/Samples/Layers.h#L87) assertion, because my object layers are laid out completely differently to how they're done in `Samples`.

How were you able to load the snapshot from #587?

jrouwe:
What I did is to set ```LoadSnapshotTest::sOverrideLayers = true```. You can do that through the menu too, but it's not obvious so I'll push a change in a minute to autodetect if there are invalid layers.

This does change the behavior, but so far the repro cases that we've gotten are mostly a single object colliding with another object so I think it will still be very useful to extract the geometry in this way.

Another thing that will probably fall over at some point is if we get a scene which relies on behavior in ```JoltGroupFilterRID``` but I'll cross that bridge when we get there.

src/shapes/jolt_shape_impl_3d.cpp
```diff
@@ -190,7 +190,65 @@ JPH::ShapeRefC JoltShapeImpl3D::without_custom_shapes(const JPH::Shape* p_shape)
...
+		}
+
+		case JPH::EShapeSubType::StaticCompound: {
+			const auto* shape = static_cast<const JPH::StaticCompoundShape*>(p_shape);
+
+			JPH::StaticCompoundShapeSettings settings;
+
+			for (const auto& sub_shape : shape->GetSubShapes()) {
+				settings.AddShape(
+					sub_shape.GetPositionCOM(),
```
jrouwe, line 207: ```sub_shape.GetPositionCOM()```

needs to be

```shape->GetCenterOfMass() + sub_shape.GetPositionCOM() - sub_shape.GetRotation() * sub_shape.mShape->GetCenterOfMass()```

because SubShapeSettings takes the model space position and internally converts to COM space.

File: src/shapes/jolt_shape_impl_3d.cpp [hunk 3]
```diff
@@ -190,7 +190,65 @@ JPH::ShapeRefC JoltShapeImpl3D::without_custom_shapes(const JPH::Shape* p_shape)
...
+		}
+
+		case JPH::EShapeSubType::MutableCompound: {
+			const auto* shape = static_cast<const JPH::MutableCompoundShape*>(p_shape);
+
+			JPH::MutableCompoundShapeSettings settings;
+
+			for (const auto& sub_shape : shape->GetSubShapes()) {
+				settings.AddShape(
+					sub_shape.GetPositionCOM(),
```
jrouwe, line 234: Same here.

src/shapes/jolt_shape_impl_3d.cpp
mihe: Ah, thank you! The moment I typed that I thought to myself that I should look into what the `COM` part was all about, but I was in too much of a hurry and forgot.

jrouwe:
I submitted the improvements in https://github.com/jrouwe/JoltPhysics/pull/712

mihe:
> Another thing that will probably fall over at some point is if we get a scene which relies on behavior in `JoltGroupFilterRID` but I'll cross that bridge when we get there.

I can confirm that things go kaboom in `GroupFilter::sRestoreFromBinaryState` when you load a snapshot with pretty much any constraints in them, since the joints in Godot filter collisions between the bodies by default, which in turn is powered by `JoltGroupFilterRID`.

jrouwe:
What you can do is set the 2nd bool in SaveBinaryState to false to skip saving the group filters. I'll make something to reconstruct them from the constraints when loading using Jolt's own primitive. That should work in most cases.

PR #599, opened 2023-09-22
Renamed "Dump Snapshots" to "Dump Debug Snapshots"

mihe:
This changes the "Dump Snapshots" menu option (as well as accompanying server methods) to be referred to as "Dump Debug Snapshots" instead.

I figured this better communicates that it's not really a useful general-purpose feature, and more of a debugging/development feature. I suppose it also opens up for some hypothetical future `space_dump_snapshot` method that doesn't filter out the custom shapes, if such a thing would happen to be useful.

CC: @jrouwe


jrouwe:
Sure, no problem!

PR #613, opened 2023-09-26
Fixed issue with `Area3D` picking up speculative contacts

### LINKED ISSUE #582: `Area3D` detects overlaps slightly outside of its collision shapes
Hi, I am making a bowling game, and after I switched to Godot Jolt, my game crashes when my level starts. The error is coming from an On Area3D Entered sensor I had set up, it is meant to detect if a bowling pin is still standing or not. The game only crashes if it starts with the pin already inside of the sensor.

The error shown is:
 "Invalid get index 'pinID' (on base: 'Node3D(pin_setter.gd)')."
 "0 - res://Meshes/Global/pin_setter.gd:71 - at function:_on_pin_up_sensor_area_entered"
 
 If I change it so the game starts without the bowling pins inside of the sensor, it runs fine.

mihe:
Fixes #582 (for real this time).

This changes the evaluation of area overlaps that happens in the contact listener to dismiss any contacts that aren't actually the result of overlapping shapes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Breaking changes are denoted with ⚠️.
   its transform relative to the joint.
 - Fixed issue where the `total_gravity` property on `PhysicsDirectBodyState3D` would always return a
   zero vector for kinematic bodies.
+- Fixed issue with `Area3D` detecting overlaps slightly outside of its collision shapes.
 
 ## [0.7.0] - 2023-08-29
 
--- a/src/spaces/jolt_contact_listener_3d.cpp
+++ b/src/spaces/jolt_contact_listener_3d.cpp
@@ -238,10 +238,12 @@ bool JoltContactListener3D::_try_evaluate_area_overlap(
 		return false;
 	}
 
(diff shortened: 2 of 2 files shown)


mihe:
@jrouwe Could I ask you to take a glance at this one?

I'm finding that areas/sensors will detect one another if they're not overlapping but are still within `mSpeculativeContactDistance`, which I guess means that I should be doing what this PR does and just ignore those contacts if I'm only interested in contacts that are the result of actual overlapping?

I must admit this one took me by surprise, although I guess it makes some kind of sense, since sensors are just regular bodies after all.

jrouwe:
I think this is just a bug in Jolt, for Sensors Jolt should not use the speculative contact distance. The speculative contact distance is only used when adding contact constraints, but that doesn't happen when you have a Sensor.

Let me see if I can create a fix (it's a bit tricky because we set the speculative contact distance before the OnContactAdded callback which can turn a body into a Sensor).

B.t.w. the change looks good, so if you want a fix now you can submit it and later roll it back.

mihe:
Great, thank you for taking a look. I'll go ahead and merge this then and roll it back if/when you get a proper fix in.

PR #620, opened 2023-09-29
Fixed small bug in physics update error handling

jrouwe:
EPhysicsUpdateError is a bitfield. Multiple errors can trigger in the same update.

### DIFF
--- a/src/spaces/jolt_space_3d.cpp
+++ b/src/spaces/jolt_space_3d.cpp
@@ -89,8 +89,12 @@ void JoltSpace3D::step(float p_step) {
 
 	_pre_step(p_step);
 
-	JPH::EPhysicsUpdateError error = physics_system->Update(p_step, 1, temp_allocator, job_system);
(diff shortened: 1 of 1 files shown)


mihe:
That's not good. I'm not sure how I missed that.

mihe:
I took the liberty of fixing the formatting issues, and lengthened the variable name just to get a consistent linebreak in all the if-statements. I'm not loving the linebreaks that clang-format decided to make in this case, but I can't really think of any creative ways to make it better. Making code look nice with clang-format is a daily struggle, and I really wish there was a better alternative out there.

I also made the variable `const`, which should ideally be caught by clang-tidy, but that particular clang-tidy rule doubles its already incredibly long run-time, so I have to run without it.

mihe:
Thank you!

jrouwe:
It's totally fine to fix formatting/style issues (I would do the same). I ran clang format myself, but it put the ```{``` on the wrong line compared to the rest of your code so I edited it by hand afterwards thinking that that was not going to be acceptable 😄.

PR #664, opened 2023-11-05
Added support for iOS and Android

### LINKED ISSUE #353: Add support for iOS
Title says it all, really. It would be a shame not to support a major platform like iOS when both Godot and Jolt supports it.

### LINKED ISSUE #354: Add support for Android
Title says it all, really. It would be a shame not to support a major platform like Android when both Godot and Jolt supports it.

mihe:
Fixes #353.
Fixes #354.

This adds (long overdue) support for building for Android (ARM64/ARM32/x64/x86) and iOS.

Note that due to limitations in GDExtension the iOS build does not support running in the iOS simulator.


jbromberg:
🐐

PR #683, opened 2023-11-14
Make cast_motion return whether or not it hit anything.

EIREXE:
Not strictly necessary since the bool isn't exposed to godot's scripting, but in my case I'm building godot-jolt as an engine module with a set of very funky workarounds. ( https://github.com/EIRTeam/godot-jolt/ ) so it would be nice if the behavior matches godotphysics

### DIFF
--- a/src/spaces/jolt_physics_direct_space_state_3d.cpp
+++ b/src/spaces/jolt_physics_direct_space_state_3d.cpp
@@ -243,7 +243,7 @@ bool JoltPhysicsDirectSpaceState3D::_cast_motion(
 	const JoltQueryFilter3D
 		query_filter(*this, p_collision_mask, p_collide_with_bodies, p_collide_with_areas);
 
-	_cast_motion_impl(
+	return _cast_motion_impl(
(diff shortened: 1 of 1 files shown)


mihe:
> so it would be nice if the behavior matches godotphysics

This was in fact changed in #466 to always return `true` specifically to match Godot Physics, as Godot Physics always returns `true` as well, as you can see [here](https://github.com/godotengine/godot/blob/c455cb65550c19afbbf657a1dbf9af2ca3b603da/servers/physics_3d/godot_space_3d.cpp#L379).

EIREXE:
> > so it would be nice if the behavior matches godotphysics
> 
> This was in fact changed in #466 to always return `true` specifically to match Godot Physics, as Godot Physics always returns `true` as well, as you can see [here](https://github.com/godotengine/godot/blob/c455cb65550c19afbbf657a1dbf9af2ca3b603da/servers/physics_3d/godot_space_3d.cpp#L379).

oh i see, wild, thanks either way

PR #710, opened 2023-12-07
Add minimum iOS version to Info.plist template

jbromberg:
When creating a release build of an iOS app containing this extension, I got errors stating that the `MinimumOSVersion` key was missing in `Info.plist`.

This PR adds that key and sets its value to iOS 15.0. The version could potentially be lower but I haven't tested it and most people are on 16 or 17 anyway.

### DIFF
--- a/cmake/templates/info.plist.in
+++ b/cmake/templates/info.plist.in
@@ -24,5 +24,7 @@
 	<string>FMWK</string>
 	<key>CSResourcesFileMapped</key>
 	<true/>
+	<key>MinimumOSVersion</key>
+	<string>15.0</string>
 </dict>
 </plist>


mihe:
Does adding that `MinimumOSVersion` entry (and only that entry) to the framework's `Info.plist` file allow you to successfully create a release build?

jbromberg:
Yes, that was all that was missing for me to create a release build.

In my case I have a paid Apple developer account and Xcode is set to automatically manage app signing. Not sure if that's the issue you're referring to.

mihe:
Right, I guess I was missing that App Store Connect thing.

I noticed that Godot is also adding the `DTPlatformName` key to its automatically [generated](https://github.com/godotengine/godot/blob/d5ad37afcd44f4ba953e3473b5ab3afb18068134/platform/ios/export/export_plugin.cpp#L1180-L1203) `Info.plist` file, so I've gone ahead and created a new PR (#712) that will make a separate `Info.plist` template for iOS instead. This will also make use of the existing `CMAKE_OSX_DEPLOYMENT_TARGET` variable, which already serves to specify the minimum target version that the library is compiled against, and can be found/changed here:

https://github.com/godot-jolt/godot-jolt/blob/1a4569d0517a431c31e8563f001cea0b27a3f796/cmake/toolchains/ios/Common.cmake#L2-L3

The choice of targeting iOS 12 is just copied from what Godot itself targets. While it obviously compiles successfully I have no idea whether it actually runs there or not.

jbromberg:
Makes sense. Thanks for working on it!

Issue #734, opened 2023-12-28
Collisions between static and locked bodies causes NaNs


Jo-Jox:
Hi, I encountered a very weird bug on Godot 4.2 .NET version with Jolt physics. I have a rotating rectangle with all linear axes locked and both angular x and angular y locked. I added a constant torque value of 100 in the inspector and ran the project. In godot without jolt, it spins and can be interacted with. With jolt, it does not spin and touching it will teleport you somewhere and cause 30 of those errors :

 E 0:00:10:0770   instance_set_transform: Condition "!v.is_finite()" is true.
  <C++ Source>   servers/rendering/renderer_scene_cull.cpp:922 @ instance_set_transform()
  
  Is there anything I did wrong ? Thanks a lot.

jrouwe:
Hello,

@Jo-Jox: Are you able to provide a minimal repro? I wasn't able to reproduce this with my own sample: [example.zip](https://github.com/godot-jolt/godot-jolt/files/13804829/example.zip) (note I didn't include the 'addons' folder in the zip, you need to copy it from elsewhere). Also, is your constant torque on the Z axis? (X or Y should not result in any rotation since the axis are locked)

Note to @mihe: I'm not clearing any locked components in Body::AddForce/AddTorque (or any of the set velocity functions) because the assumption is that the user will not try to force a velocity on a locked axis. So, while the inertia tensor should be zero in the locked axis direction, there could be some residual value that ends up in a locked axis direction. To prevent this, these lines should be changed to clear components for locked axis:

https://github.com/godot-jolt/godot-jolt/blob/7f6fe9fd3ec0dc8045b84d3d04dbdc20124dea59/src/objects/jolt_body_impl_3d.cpp#L1162-L1163

For translation there is ```MotionProperties::LockTranslation```. I will make a function for rotation too.

Jo-Jox:
I have removed everything else from my scene to only left the weird behaving objects so you can see what I mean. The weird thing is when creating a whole new project the issue doesn't seem to bee there anymore... There will be some errors as a deleted a whole bunch of things but the issue is still testable. I removed the jolt folders due to size so it must be installed again.
[Example.zip](https://github.com/godot-jolt/godot-jolt/files/13805063/Example.zip)

jrouwe:
I've further minimized the example: [Example2.zip](https://github.com/godot-jolt/godot-jolt/files/13805137/Example2.zip)

[...]

PR #742, opened 2024-01-03
Added clearing of locked axes when changing velocities/forces

mihe:
As talked about in #734, Jolt assumes that the user will not apply forces/impulses/velocities on any currently locked axes, so this PR makes sure to explicitly clear those to avoid any problems with that.

### DIFF
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -354,15 +354,18 @@ void JoltBodyImpl3D::set_angular_velocity(const Vector3& p_velocity) {
 		return;
 	}
 
+	// We need to clear locked angular axes since Jolt currently only clears linear axes
+	const Vector3 locked_velocity = _stop_locked_angular_axes(p_velocity);
+
 	if (space == nullptr) {
-		jolt_settings->mAngularVelocity = to_jolt(p_velocity);
+		jolt_settings->mAngularVelocity = to_jolt(locked_velocity);
 		return;
 	}
 
 	const JoltWritableBody3D body = space->write_body(jolt_id);
 	ERR_FAIL_COND(body.is_invalid());
 
-	body->GetMotionPropertiesUnchecked()->SetAngularVelocityClamped(to_jolt(p_velocity));
+	body->GetMotionPropertiesUnchecked()->SetAngularVelocityClamped(to_jolt(locked_velocity));
 }
 
 void JoltBodyImpl3D::set_axis_velocity(const Vector3& p_axis_velocity) {
@@ -537,7 +540,17 @@ void JoltBodyImpl3D::apply_force(const Vector3& p_force, const Vector3& p_positi
 	const JoltWritableBody3D body = space->write_body(jolt_id);
 	ERR_FAIL_COND(body.is_invalid());
 
-	body->AddForce(to_jolt(p_force), body->GetPosition() + to_jolt(p_position));
+	// We need to manually apply any induced torque despite there being an overload of `AddForce`
+	// that does this for us, since Jolt doesn't currently clear locked angular axes.
+
+	const Vector3 body_pos = to_godot(body->GetPosition());
+	const Vector3 body_com = to_godot(body->GetCenterOfMassPosition());
+	const Vector3 torque = (body_pos + p_position - body_com).cross(p_force);
+
+	const Vector3 locked_torque = _stop_locked_angular_axes(torque);
+
+	body->AddForce(to_jolt(p_force));
+	body->AddTorque(to_jolt(locked_torque));
 
 	_motion_changed();
 }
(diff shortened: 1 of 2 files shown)


mihe:
@jrouwe I must admit I don't love how this all turned out in practice, assuming I did things correctly of course.

For things like the non-central forces I'm now forced to apply the resulting torque myself just so that I can properly affect any unlocked angular axes, in the case where all linear axes are locked but not all angular ones are.

For the non-central impulses I opted to just clear the resulting velocities instead, which is fine I guess but does make for an odd discrepancy between it and the non-central `apply_force`.

(I opted not to use your `LockTranslation` and `LockAngular` because I need these methods even when there is no `JPH::Body` as well as for kinematic bodies which don't use `EAllowedDOFs`.)

I realize we might be straying into Godot-specific territory here, but this coupled with the need to transform the inertia tensor back-and-forth is making this `EAllowedDOFs` thing feel a bit cumbersome from a Jolt user perspective. I'm not sure what the alternative is exactly. Obviously saddling every user of Jolt with the cost of clearing the applicable axes for every `AddForce`, `Set*Velocity`, etc. is not ideal, but maybe some other ideas spring to mind for you.

jrouwe:
> this coupled with the need to transform the inertia tensor back-and-forth (in order to lock in global space) is making this EAllowedDOFs thing feel a bit cumbersome

You mean this code?

https://github.com/mihe/godot-jolt/blob/1cb2c55bcf2bfa71c8787dbb6206ec49d5fcc6ac/src/objects/jolt_body_impl_3d.cpp#L1348-L1393

I had made a note of that but didn't get around to making something in Jolt for it yet.

Thinking about it... doesn't this change break the angular locking when the rotation axis of the body are not axis aligned?

E.g. this code:

https://github.com/mihe/godot-jolt/blob/1cb2c55bcf2bfa71c8787dbb6206ec49d5fcc6ac/src/objects/jolt_body_impl_3d.cpp#L359

Will happily reset the world space X axis if X is locked where in fact it should reset X in local space of the body?

In any case I could fix some of the functions inside Jolt, e.g. AddForce/AddTorque, the main reason I haven't is that especially the Set*Velocity functions are called a ton and I didn't want to make them any slower. But there are other people who apparently ran into the same issue and who opted to do the clearing inside Jolt, see [HypeHype's change](https://github.com/Frogmind/JoltPhysics/commit/ea9ba52fa012e1d774e60319768b0842bfb4e9c7#diff-486dfee402ec3973a92c19eeb8135e008c6e9a80747fc625e8d8b71eea5dbbe4). 

If I were to reset the angular components then I will be breaking the code in the same way as your change is now though as Jolt is unaware of the locking in body space.

jrouwe:
Thinking about this a bit more: I'll add a LockTranslation to SetLinearVelocity(Clamped). It was called a lot less than I thought so it should be fine. This will ensure that you don't need to reset the linear components when calling AddImpulse and AddForce too because that will be handled internally.

This leaves the angular locking problem, and I'm guessing the only way that this can be fixed (besides doing it in godot-jolt) is by introducing a mode inside Jolt that allows rotation locking in local space so that the LockAngular function can reset in local space.

jrouwe:
Edit: I'm an idiot and misremembered how I implemented things. Jolt is locking rotation in local space, not godot. So your code is correct.

mihe:
Apologies for the late reply.

> You mean this code?

Yes, exactly.

> But there are other people who apparently ran into the same issue and who opted to do the clearing inside Jolt, see [HypeHype's change](https://github.com/Frogmind/JoltPhysics/commit/ea9ba52fa012e1d774e60319768b0842bfb4e9c7#diff-486dfee402ec3973a92c19eeb8135e008c6e9a80747fc625e8d8b71eea5dbbe4).

I assume you meant to link [this change](https://github.com/Frogmind/JoltPhysics/commit/b15169e6d9a1f8519a51198de32b0e4f087719ce)?

> Thinking about this a bit more: I'll add a LockTranslation to SetLinearVelocity(Clamped). It was called a lot less than I thought so it should be fine. This will ensure that you don't need to reset the linear components when calling AddImpulse and AddForce too because that will be handled internally.

Thank you. That did get rid of some of the code.

> This leaves the angular locking problem, and I'm guessing the only way that this can be fixed (besides doing it in godot-jolt) is by introducing a mode inside Jolt that allows rotation locking in world space.

That sounds about right, and would be much appreciated.

jrouwe:
> I assume you meant to link [this change](https://github.com/Frogmind/JoltPhysics/commit/b15169e6d9a1f8519a51198de32b0e4f087719ce)?

Yes, I think they actually go together. The one I linked removes the clearing of the inertia tensor components, the one you reference clears the components.

> That sounds about right, and would be much appreciated.

Ok, I'll see what I can do.

Issue #357, opened 2023-05-18
Memory usage is not reported anywhere


mihe:
Due to Jolt needing aligned allocations we are currently unable to use the [default allocators](https://github.com/godotengine/godot-cpp/blob/1c18413de00f1a6265b2b6c30175b2f6a434b574/src/core/memory.cpp) that godot-cpp provides. This also means that Godot is unaware of any memory that gets allocated by Jolt or Godot Jolt and as such doesn't reflect it in its performance monitors.

One way to alleviate this is by adding custom performance monitors that track the memory used by Jolt.

There's a great example of this in the [FMOD for Godot](https://github.com/alessandrofama/fmod-for-godot/blob/22fccf66bed93c437a0b0a77e9e3b5118ccec964/addons/FMOD/native/src/utils/debug_monitors.cpp) extension.

mihe:
This is unlikely to be worked on at this point, as it's looking quite likely that this extension will be migrated over to Godot as a first-party module in a not too distant future, so I'll go ahead and close this.

PR #760, opened 2024-01-26
Attempt to add memory monitors

Kubulambula:
I tried to implement a memory monitor described in this [Issue](https://github.com/godot-jolt/godot-jolt/issues/357).

It uses custom allocators to track the memory usage in debug builds and provides a performance monitor to the Godot editor under a `Jolt Physics` category. There are monitors for current and maximal memory usage formatted to MiB. 

There could be a ProjectSetting for the formatting, but I don't really know about that. The memory is reported as a decimal value, but Godot's performance monitor does not format custom monitor values. It could be changed to report the value as an integer, but at the expense of monitor granularity.

image
_How it looks in the editor_

This might not be an ideal implementation whatsoever, so any feedback or critique would be highly appreciated as I don't have that much experience with C++.

### DIFF
--- a/src/servers/jolt_globals.cpp
+++ b/src/servers/jolt_globals.cpp
@@ -6,27 +6,90 @@
 #include "shapes/jolt_custom_ray_shape.hpp"
 #include "shapes/jolt_custom_user_data_shape.hpp"
 
-#ifdef GDJ_USE_MIMALLOC
+#ifdef GDJ_CONFIG_EDITOR
+#include "servers/jolt_physics_server_3d.hpp"
+#include <unordered_map>
 
+std::unordered_map<void*, size_t> alloc_size_map;
+#endif // GDJ_CONFIG_EDITOR
+
+#ifdef GDJ_USE_MIMALLOC
 #include <mimalloc-new-delete.h>
+#endif // GDJ_USE_MIMALLOC
 
 void* jolt_alloc(size_t p_size) {
-	return mi_malloc(p_size);
+#ifdef GDJ_USE_MIMALLOC
+	void* block = mi_malloc(p_size);
+#else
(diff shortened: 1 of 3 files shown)


mihe:
I appreciate the effort you've put it into this, but there are a number of issues with this implementation:

1. The tracking of allocations through a `unordered_map` is going to be quite slow for something as core as allocation of memory. Granted, most allocations Jolt itself does happen as part of its temp/stack/bump allocator, so won't be affected by this, but the rest of the extension will be.
2. The tracking doesn't take into considering that allocations can happen from multiple threads at the same time, so will be subject to race conditions.
3. The tracking is initialized even in what I (somewhat confusingly) refer to as editor builds, meaning what Godot calls debug builds.
5. Minor stylistic issues here and there.

Frankly I would rather utilize mimalloc for this, when applicable. It keeps track of all of this internally already. From what I can tell it's possible to use the `mi_heap_visit_blocks` function to get at this information, but I'm not entirely sure how costly that would be to call every time the performance monitor updates. Ideally they would've exposed the stats directly, but for some reason they chose not to.

Another issue of course what you do when mimalloc isn't used, like with iOS and Android builds, in which case I'd rather see something like what Godot itself is doing, where they overallocate by a certain amount and put the size at the beginning of every allocation, which you can then easily and quickly read from when freeing. Care has to taken when doing this for aligned allocations though, since the resulting pointer returned has to be similarly aligned.

mihe:
Again, I appreciate the effort you put into this, but I'm gonna go ahead and close this.

It's looking likely that this extension will be integrated into Godot as an official engine module in a (hopefully) not too distant future, in which case this extension would likely be archived, or at least put into some kind of maintenance mode, meaning any further effort invested into this particular PR would go wasted.

PR #765, opened 2024-01-31
Added support for `SoftBody3D`

### LINKED ISSUE #503: Support for `SoftBody3D`
I've been working on soft body support and I'm trying to prioritize features that will be used by godot. I'm creating this ticket to let you know of my progress, the kind of things I'm running into and to have a place to discuss the feature.

First of all I have a difficult time to narrow down what it exactly is that godot supports. I've created a test scene:

[SoftBody.zip](https://github.com/godot-jolt/godot-jolt/files/12136548/SoftBody.zip)

As far as I can see there no collision detection implemented between soft bodies, but it is also quite possible that I don't understand the layer/mask pairs and that it is possible to get the 2 to collide. Am I doing something wrong?

https://github.com/godot-jolt/godot-jolt/assets/1621693/7c542660-ce57-453d-b1da-849fa47e62c9

B.t.w. I'm pinning one of the vertices of each shape, because if I don't this is what I get:

https://github.com/godot-jolt/godot-jolt/assets/1621693/e5a10bcb-11a7-49ee-9548-ea2e5c6f2c8f

The rigid body sphere collides with the soft bodies, but while playing around with this I found the collision response very unreliable. Often the sphere passes through the soft body without being affected by it. I also had to export my own sphere from blender because the default sphere mesh (or any other default godot mesh) has a seam so will deflate immediately. This is a movie of when it works:

https://github.com/godot-jolt/godot-jolt/assets/1621693/5ebd6509-f7cb-4d71-9a4f-01fb6aad9ed0

Finally while implementing 'pressure' I ran into what I think is a bug in godot. The force on the vertices of the soft body is calculated by ```force = pressure_constant * area / volume``` but I think the area calculation is incorrect:

https://github.com/godotengine/godot/blob/6588a4a29af1621086feac0117d5d4d37af957fd/servers/physics_3d/godot_soft_body_3d.cpp#L255-L261

In this loop it distributes the area of the faces of the soft body to the vertices, which will later be used to distribute the pressure force. Consider this simple octahedron:

image

The loop will add the area A to the vertex v 4 times and then divide by 4 again in the next code block. The same will happen for all other vertices in the shape, this means that the total area is 6 * A while it in fact should be 8 * A.

I'm obviously not going to copy this behavior, which means that the pressure coefficient is going to behave different in Jolt than it will in godot (and it will depend on the topology of the shape how different it is).

mihe:
Fixes #503.

This adds support for the `SoftBody3D` node.

This also brings with it a number of refactorings:

1. `JoltObjectImpl3D` has partially been split into a new `JoltShapedObjectImpl3D`.
2. `JoltBodyImpl3D` and `JoltAreaImpl3D` now inherit from `JoltShapedObjectImpl3D` instead.
3. Creation of Jolt bodies was simplified/flattened a bit.
4. Much of the group filter is now done in the object's `can_interact_with` method instead.
5. Global gravity is now a tad more complicated, due to needing to use Jolt's actual gravity for soft bodies.

(This will need a lot of testing before it can be merged.)

### DIFF
--- a/src/objects/jolt_soft_body_impl_3d.cpp
+++ b/src/objects/jolt_soft_body_impl_3d.cpp
@@ -76,7 +76,7 @@ void JoltSoftBodyImpl3D::set_mesh(const RID& p_mesh) {
 void JoltSoftBodyImpl3D::set_simulation_precision(int32_t p_precision) {
 	QUIET_FAIL_COND(simulation_precision == p_precision);
 
-	simulation_precision = p_precision;
+	simulation_precision = MAX(p_precision, 0);
 
 	_simulation_precision_changed();
 }
@@ -84,7 +84,7 @@ void JoltSoftBodyImpl3D::set_simulation_precision(int32_t p_precision) {
 void JoltSoftBodyImpl3D::set_mass(float p_mass) {
 	QUIET_FAIL_COND(mass == p_mass);
 
-	mass = p_mass;
+	mass = MAX(p_mass, 0.0f);
 
 	_mass_changed();
 }
@@ -96,13 +96,13 @@ float JoltSoftBodyImpl3D::get_stiffness_coefficient() const {
 void JoltSoftBodyImpl3D::set_stiffness_coefficient(float p_coefficient) {
 	QUIET_FAIL_COND(stiffness_coefficient == p_coefficient);
 
-	stiffness_coefficient = clamp(p_coefficient, 0.0f, 1.0f);
+	stiffness_coefficient = CLAMP(p_coefficient, 0.0f, 1.0f);
 }
 
 void JoltSoftBodyImpl3D::set_pressure(float p_pressure) {
 	QUIET_FAIL_COND(pressure == p_pressure);
 
-	pressure = p_pressure;
+	pressure = MAX(p_pressure, 0.0f);
 
 	_pressure_changed();
 }
@@ -110,7 +110,7 @@ void JoltSoftBodyImpl3D::set_pressure(float p_pressure) {
 void JoltSoftBodyImpl3D::set_linear_damping(float p_damping) {
 	QUIET_FAIL_COND(linear_damping == p_damping);
 
-	linear_damping = p_damping;
+	linear_damping = MAX(p_damping, 0.0f);
 
 	_damping_changed();
 }
@@ -567,7 +567,7 @@ bool JoltSoftBodyImpl3D::_ref_shared_data() {
 		// HACK(mihe): Since Godot's stiffness is input as a coefficient between 0 and 1, and Jolt
 		// uses actual stiffness for its edge constraints, we crudely map one to the other with an
 		// arbitrary constant.
-		const float stiffness = max(stiffness_coefficient * 100'000.0f, 0.00001f);
+		const float stiffness = MAX(stiffness_coefficient * 100'000.0f, 0.00001f);
 		const float inverse_stiffness = 1.0f / stiffness;
 
 		SymmetricBitTable marked_edges((int32_t)physics_vertices.size());
--- a/src/objects/jolt_soft_body_impl_3d.cpp
+++ b/src/objects/jolt_soft_body_impl_3d.cpp
@@ -254,32 +254,48 @@ void JoltSoftBodyImpl3D::update_rendering_server(
 	const JoltReadableBody3D body = space->read_body(jolt_id);
 	ERR_FAIL_COND(body.is_invalid());
 
-	const Vector3 center_of_mass = to_godot(body->GetCenterOfMassPosition());
-
 	const auto& motion_properties = static_cast<const JPH::SoftBodyMotionProperties&>(
 		*body->GetMotionPropertiesUnchecked()
 	);
 
-	const JPH::Array<JPH::SoftBodyVertex>& physics_vertices = motion_properties.GetVertices();
+	using SoftBodyVertex = JPH::SoftBodyMotionProperties::Vertex;
+	using SoftBodyFace = JPH::SoftBodyMotionProperties::Face;
 
-	Vector3 face[3];
+	const JPH::Array<SoftBodyVertex>& physics_vertices = motion_properties.GetVertices();
+	const JPH::Array<SoftBodyFace>& physics_faces = motion_properties.GetFaces();
 
-	for (int32_t i = 0; i < shared->mesh_to_physics.size(); i += 3) {
-		for (int32_t j = 0; j < 3; ++j) {
-			const int32_t mesh_index = i + j;
-			const auto physics_index = (size_t)shared->mesh_to_physics[mesh_index];
-			const Vector3 local_vertex = to_godot(physics_vertices[physics_index].mPosition);
-			face[j] = center_of_mass + local_vertex;
-			p_rendering_server_handler->set_vertex(mesh_index, face[j]);
-		}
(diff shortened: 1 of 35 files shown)


src/objects/jolt_soft_body_impl_3d.cpp
```diff
@@ -0,0 +1,744 @@
...
+				(JPH::uint32)physics_face[1],
+				(JPH::uint32)physics_face[0]
+			);
+		}
+
+		// HACK(mihe): Since Godot's stiffness is input as a coefficient between 0 and 1, and Jolt
+		// uses actual stiffness for its edge constraints, we crudely map one to the other with an
+		// arbitrary constant.
+		const float stiffness = MAX(stiffness_coefficient * 100'000.0f, 0.00001f);
+		const float inverse_stiffness = 1.0f / stiffness;
```
mihe, line 571: @jrouwe I'm curious if you have any better idea for how to deal with `mCompliance`.

The `linear_stiffness` property in Godot is sadly not actual stiffness but instead a fraction/coefficient between 0 and 1, and I'm running out of ideas for how to best map from this to actual stiffness, so I just settled for this naive/crude linear mapping.

In Godot Physics this particular property does nothing, as it turns out, as it's only ever incorporated into one of the "link constants" (see [here](https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/servers/physics_3d/godot_soft_body_3d.cpp#L280)) but that constant is never actually used anywhere.

In Bullet however, from where I'm assuming this was copied, this "link constant" does do something, but I'm struggling to figure out exactly how it's different from Jolt, and whether there's a way to map between the two. It doesn't seem to be affected by mass at all, from what I can tell, as the constraints seem to act about the same no matter how much mass I load on.

Given that this property isn't actually used in Godot Physics I wouldn't be entirely against just not letting this be configurable (which would also make the caching of the shared soft body settings easier) and instead either set it to be some constant or calculate it automatically somehow. I tried a couple of variations based on mass, gravity and the rest length of the edge, but I couldn't really find anything that made much sense, nor felt intuitive.

Maybe just hardcoding it to `0.000001f` is the better play?

jrouwe: Are you sure it isn't used? To me it looks like it is used here:

https://github.com/godotengine/godot/blob/9adb7c7d130c6d4eb0e80b92d6eebd71fac3384d/servers/physics_3d/godot_soft_body_3d.cpp#L1104

(it is used in link.c2 as well but that constant doesn't seem to be used)

mihe: You're absolutely right. I'm not sure why I stopped looking after the first occurrence. That's pretty much the same as with Bullet then: https://github.com/godotengine/godot/blob/44cb43b5e73a4b40ae20cc5d434d3e70cb0672c0/thirdparty/bullet/BulletSoftBody/btSoftBody.cpp#L4009

That makes the hardcoded value less appealing I guess.

Either way, would you happen to have any suggestions for how to map one to the other?

jrouwe: If I compare the formulas in Bullet and GodotJolt then I think they're pretty much the same. The update of an edge constraint boils down to:

image

(and something similar for delta_p2)

Where p1 and p2 are the positions of the particles, m1 and m2 the masses, rest_len is the rest length between the particles, len the current length (|p2 - p1|). 

This formula strikes me as very odd since rest_len^2 doesn't have the same units as len so you cannot subtract them. I don't recognize the formula, I would expect something like eq 10 of https://matthias-research.github.io/pages/publications/posBasedDyn.pdf.

If eq 10 was the intention then a linear_stiffness (with which we would multiply the right hand side of eq 10) between 0 and 1 starts making sense: If it is 0 then nothing of the position is corrected, if it is 1 then the edge length is fully updated to rest_len. So a linear_stiffness of 1 would map to a compliance of 0 and a linear stiffness of 0 would map to a compliance of infinite. Problem is that godot isn't using eq 10 so I'm not sure what to make of this.

jrouwe: For reference, this is what happens in Jolt with XPBD:

image

mihe: Interesting. Thank you for the breakdown/analysis.

I'm getting the sense that this mapping will end up being somewhat arbitrary in that case. I might take some creative liberties and make it a cubic mapping then. That way it should be possible to make things loose at the lower ranges with a low mass, since the increments are 0.01, while still being able to really crank it up if need be.

mihe: On a somewhat related note, is there any chance that there could be some sort of per-body override of `mCompliance`, so that I can avoid having to use the per-edge `mCompliance` altogether? With the way things are now in Jolt, I don't see any good way of actually changing compliance after the body has been created, which is something you can do in Godot.

Recreating the body through `GetSoftBodyCreationSettings` doesn't seem to set new initial positions of all the vertices, since it just copies the settings pointer, which I guess makes sense, but that also means you can't really adjust the compliance without a lot of jank. Even if `GetSoftBodyCreationSettings`/`SoftBodySharedSettings` did support this, I would need to cache the shared settings based on not only the mesh, but stiffness as well, which becomes somewhat wasteful, given that this stiffness applies to the entire body in Godot.

jrouwe: > On a somewhat related note, is there any chance that there could be some sort of per-body override of mCompliance, so that I can avoid having to use the per-edge mCompliance altogether?

I would prefer not to do this. This would basically mean that I need to duplicate this code:

https://github.com/jrouwe/JoltPhysics/blob/a6c9f8808f78c760ed0bed6f2e6c6129e66e9203/Jolt/Physics/SoftBody/SoftBodyMotionProperties.cpp#L287-L303

with the only difference that ```e.mCompliance``` has a different value. A conditional in that place is not a good idea since it is the innermost loop and executed many thousand times per frame.

I don't think you need to recreate the body if the compliance changes. I'm assuming you create a single SoftBodySharedSettings object per soft body since I don't think godot shares anything between the soft bodies (but maybe I'm wrong considering your remark above). I think you can just loop over the edges and update the compliance and the soft body will automatically use the new values.

mihe: > I would prefer not to do this.

Fair enough. That makes sense. I figured I should check at least.

> I'm assuming you create a single SoftBodySharedSettings object per soft body since I don't think godot shares anything between the soft bodies (but maybe I'm wrong considering your remark above).

Godot does in fact share resources quite aggressively. If you duplicate a `SoftBody3D` you'll end up using the same `Mesh` resource for both of them, as well as whatever other resources it might reference. It seemed like a shame not to utilize this trait and cache Jolt's `SoftBodySharedSettings` based on this mesh resource ID.

I guess the question becomes whether the caching/sharing is worth more than the ability to change `linear_stiffness` after the `SoftBody3D` has been assigned a mesh, which is when the underlying Jolt body gets created. I have a hard time imagining that many people would mess with `linear_stiffness` in any sort of dynamic fashion, whereas many more people might benefit from the performance/memory benefits that the caching brings with it.

I'm also fairly certain that you can't actually change `linear_stiffness` dynamically in Godot Physics either, or in Bullet in Godot 3 for that matter, as they don't seem to actually update the edge constants after changing `linear_stiffness`.

Norrox:
Great work @mihe !!!!

PR #775, opened 2024-02-11
Added project setting for enhanced internal edge removal

### LINKED ISSUE #530: `move_and_collide` against triangle intersection yields odd normals
This may seem minor, but these errors add up. I was only able to discover them by walking around with my character controller. I managed to isolate the issue in an MRP. I'm sorry I dont have much else to say, No I am not technical with collision detection, I simply rely on getting 100% correct collision normals so my [custom character physics](https://github.com/Gamepro5/Custom_CharacterBody3D_Movement_System/blob/main/player/CharactedController.gd) can work.

MRP:
[addons.zip](https://github.com/godot-jolt/godot-jolt/files/12245895/addons.zip)

### LINKED ISSUE #587: Sphere shapes collide with inactive edges

### LINKED ISSUE #764: Springs and high triangle collision surfaces
Having an issue that I'm not entirely sure how to work around nor the actual cause. When using RigidBody3D with JoltGeneric6DOFJoint3D to build a vehicle, I'm experiencing bumpy response on my springs in combination with surfaces generated from Terrain3D.

Here's a video of the issue: https://youtu.be/UJ2mkj5iNHk

Here is a full project I built out to test this: https://drive.google.com/file/d/1NrZustmjPDWhwhlTJW0_hugiSTIhxISU/view?usp=sharing

I apologize if I missed something obvious or a setting I can use to fix this. I'd be glad to create a video when I'm done if it is something simple to help educate others.

Thank you

### LINKED ISSUE #771: CSG nodes yield weird results
It is possible that the way Jolt handles CSG nodes leads to poor floor recognition as shown in this issue https://github.com/mrezai/GodotStairs/issues/8

mihe:
Fixes #530.
Fixes #587.
Fixes #764.
Fixes #771.

This adds a project setting for enabling Jolt's new "enhanced internal edge removal", which helps alleviate ghost collisions.

Note that this applies not only to `RigidBody3D` but also physics queries like `move_and_collide` and `move_and_slide`, which means `CharacterBody3D` benefits from this as well.

### DIFF
--- a/src/spaces/jolt_physics_direct_space_state_3d.cpp
+++ b/src/spaces/jolt_physics_direct_space_state_3d.cpp
@@ -170,7 +170,6 @@ int32_t JoltPhysicsDirectSpaceState3D::_intersect_shape(
 	settings.mMaxSeparationDistance = (float)p_margin;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
@@ -255,7 +254,6 @@ bool JoltPhysicsDirectSpaceState3D::_cast_motion(
 	settings.mMaxSeparationDistance = (float)p_margin;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
@@ -315,7 +313,6 @@ bool JoltPhysicsDirectSpaceState3D::_collide_shape(
 	settings.mMaxSeparationDistance = (float)p_margin;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
@@ -386,7 +383,6 @@ bool JoltPhysicsDirectSpaceState3D::_rest_info(
 	settings.mMaxSeparationDistance = (float)p_margin;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
@@ -756,7 +752,6 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_recover(
 	settings.mMaxSeparationDistance = p_margin;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
@@ -863,7 +858,6 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_cast(
 	JPH::CollideShapeSettings settings;
 
 	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
 		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 	}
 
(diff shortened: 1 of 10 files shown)


src/spaces/jolt_physics_direct_space_state_3d.cpp
```diff
@@ -169,10 +169,15 @@ int32_t JoltPhysicsDirectSpaceState3D::_intersect_shape(
 	JPH::CollideShapeSettings settings;
 	settings.mMaxSeparationDistance = (float)p_margin;
 
+	if (JoltProjectSettings::use_enhanced_edge_removal()) {
+		settings.mActiveEdgeMode = JPH::EActiveEdgeMode::CollideWithAll;
+		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
+	}
```
mihe, line 175: @jrouwe Am I right in enabling `EActiveEdgeMode::CollideWithAll` when using `InternalEdgeRemovingCollector`? I assume this new collector makes the active edge detection redundant?

(The results I've seen from some preliminary testing are great by the way.)

jrouwe: It should be redundant, but I think it is still faster to leave the other on as well because of this early out:

https://github.com/jrouwe/JoltPhysics/blob/398f571b1a7c30be330bc6c5ce2e3aaffdc93916/Jolt/Physics/Collision/InternalEdgeRemovingCollector.h#L115

jrouwe: B.t.w. you may want to put this one back to default again:

https://github.com/jrouwe/JoltPhysics/blob/398f571b1a7c30be330bc6c5ce2e3aaffdc93916/Jolt/Physics/Collision/Shape/MeshShape.h#L60

mihe: Ah, gotcha, thank you. I went ahead and made it use a fixed threshold of 5 degrees when you enable this new enhanced internal edge removal. I also went ahead and enabled the setting by default.

Gamepro5:
This was quite the surprise to see this morning. I can't wait to test how this affects my projects! Thanks for not forgetting about the ghost collision issue!

Saul2022:
This look’s pretty good  and will definetly fix my issue  with jumping mechanichs on terrain though i  may be a bit late, but how does  it may affect performance?

mihe:
@Saul2022 See [this reply](https://github.com/godot-jolt/godot-jolt/issues/587#issuecomment-1902255928). It will likely vary from scene to scene, but it can apparently be faster in some cases even. I haven't done any profiling myself though. If it negatively affects performance for some reason then you can always just disable the setting I guess.

Gamepro5:
@mihe , Might be a dumb question but how do I test this feature? The last release was in January. Do I need to compile it from main or are there nightly builds?

mihe:
You need to build it yourself for now, yes. I had hoped to publish a new release by now, but I wanted to try and get a few more things in first, and things like the double-precision builds need more testing before I'm comfortable putting out the next stable release.

At the very least there will be a pre-release for Godot 4.3 up here on GitHub shortly after 4.3-beta1 is out, which should be around the corner.

Gamepro5:
Yeah it does seem like this fixed everything. Awesome work!

mihe:
Glad to hear it!

I'm guessing @jrouwe will be glad to hear that as well.

Saitodepaula:
> You need to build it yourself for now, yes. I had hoped to publish a new release by now, but I wanted to try and get a few more things in first, and things like the double-precision builds need more testing before I'm comfortable putting out the next stable release.
> 
> At the very least there will be a pre-release for Godot 4.3 up here on GitHub shortly after 4.3-beta1 is out, which should be around the corner.

With Jolt 5.0 released, any idea when a new release could be published?

mihe:
Jolt's versions don't mean a whole lot in the context of this extension. I've stayed pretty close to the head of its `master` branch since the beginning.

As mentioned in this thread already, there will be a pre-release (here on GitHub) some time around Godot 4.3-beta1 that will target Godot 4.3. Unfortunately Godot 4.3 has been delayed by quite a bit since I wrote that message, but I'd rather not release a stable release this late in the lifecycle of Godot 4.2, especially not with the magnitude of the changes in this pending release.

Godot 4.3-beta1 should (hopefully for real this time) be just around the corner, in like a week or two.

PR #780, opened 2024-02-15
Added support for double-precision

### LINKED ISSUE #355: Add support for double-precision
This would add support for builds of Godot that use double-precision floating-point numbers, enabled by building Godot with `precision=double`, as opposed to the default single-precision.

This would allow for much bigger worlds.

mihe:
Fixes #355.

This adds support for building binaries that are compatible with double-precision builds of Godot, which is done by passing `-DGDJ_DOUBLE_PRECISION=TRUE` when generating the build directory.

### DIFF
--- a/.github/actions/cmake-configure/action.yml
+++ b/.github/actions/cmake-configure/action.yml
@@ -5,6 +5,8 @@ inputs:
     required: true
   arch:
     default: null
+  double:
+    default: false
   args:
     default: ''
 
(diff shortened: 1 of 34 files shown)


cridenour:
Incredible. Already built and tested and works like a dream.

mihe:
That was quick! I haven't tested the physics queries much at all, so if you're using stuff like ray-casts and shape-casts let me know how they're working out for you.

cridenour:
Doing some more testing at 100km from origin:

Raycasts: Work great, both from node and PhysicsServer3D directly.
Shapecasts: Same as above.

That said, CharacterBody3D `move_and_slide` seems to have trouble at this distance. It did fine at 25km which made me believe doubles were working but at 100km the character is unmovable vs. GodotPhysics which shows no difference.

mihe:
Ah, that figures. I should have tested `move_and_slide` before pushing it. I'll fix it shortly.

mihe:
It turned out to be an easy fix, thankfully. (#781)

PR #781, opened 2024-02-15
Fixed precision issue in `test_body_motion`

### LINKED ISSUE #355: Add support for double-precision

mihe:
This fixes a precision truncation in `test_body_motion`, and thus `move_and_collide` and `move_and_slide` as well.


cridenour:
Works great. Can't find any more double issues in my testing!

PR #806, opened 2024-03-15
Fixed issue with shape queries not returning the full manifold

mihe:
Fixes #804.

This fixes a discrepancy with Godot Physics when using any of the following methods:

- `PhysicsDirectSpaceState3D.collide_shape`
- `PhysicsServer3D.body_test_motion`
- `PhysicsBody3D.test_move`
- `PhysicsBody3D.move_and_collide`
- `CharacterBody3D.move_and_slide`

... where until now you would at most only ever get a single contact point per shape you were colliding with. Now instead you will get the full contact manifold, assuming of course that the `max_collisions` or `max_results` parameter allows for it.

### DIFF
--- a/src/spaces/jolt_physics_direct_space_state_3d.cpp
+++ b/src/spaces/jolt_physics_direct_space_state_3d.cpp
@@ -311,10 +311,7 @@ bool JoltPhysicsDirectSpaceState3D::_collide_shape(
 
 	JPH::CollideShapeSettings settings;
 	settings.mMaxSeparationDistance = (float)p_margin;
-
-	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
-	}
+	settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 
 	const Vector3& base_offset = transform_com.origin;
 
@@ -339,19 +336,46 @@ bool JoltPhysicsDirectSpaceState3D::_collide_shape(
 		return false;
 	}
 
-	auto* results = static_cast<Vector3*>(p_results);
+	auto* points = static_cast<Vector3*>(p_results);
 
-	*p_result_count = collector.get_hit_count();
+	const int32_t max_points = p_max_results * 2;
 
-	for (int32_t i = 0; i < *p_result_count; ++i) {
+	int32_t point_count = 0;
+
+	for (int32_t i = 0; i < collector.get_hit_count(); ++i) {
 		const JPH::CollideShapeResult& hit = collector.get_hit(i);
 
 		const Vector3 penetration_axis = to_godot(hit.mPenetrationAxis.Normalized());
 		const Vector3 margin_offset = penetration_axis * (float)p_margin;
 
-		*results++ = base_offset + to_godot(hit.mContactPointOn1) + margin_offset;
-		*results++ = base_offset + to_godot(hit.mContactPointOn2);
+		JPH::ContactPoints contact_points1, contact_points2;
+
+		_generate_manifold(
+			hit,
+			settings,
+			contact_points1,
+			contact_points2
+#ifdef JPH_DEBUG_RENDERER
+			,
+			to_jolt_r(base_offset)
+#endif // JPH_DEBUG_RENDERER
+		);
+
+		for (JPH::uint j = 0; j < contact_points1.size(); ++j) {
+			points[point_count++] = base_offset + to_godot(contact_points1[j]) + margin_offset;
+			points[point_count++] = base_offset + to_godot(contact_points2[j]);
+
+			if (point_count >= max_points) {
+				break;
+			}
+		}
+
+		if (point_count >= max_points) {
+			break;
+		}
 	}
+
+	*p_result_count = point_count / 2;
 
 	return true;
 }
@@ -930,10 +954,7 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 
 	JPH::CollideShapeSettings settings;
 	settings.mMaxSeparationDistance = p_margin;
-
-	if (JoltProjectSettings::use_enhanced_edge_removal()) {
-		settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
-	}
+	settings.mCollectFacesMode = JPH::ECollectFacesMode::CollectFaces;
 
 	const Vector3& base_offset = transform_com.origin;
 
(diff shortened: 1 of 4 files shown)


src/spaces/jolt_physics_direct_space_state_3d.cpp
```diff
@@ -983,28 +1021,73 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
...
+	);
+
+	if (p_contact_points1.size() > 4) {
+		JPH::PruneContactPoints(
+			p_hit.mPenetrationAxis,
+			p_contact_points1,
+			p_contact_points2,
+			p_center_of_mass
+		);
+	}
```
mihe, line 1092: @jrouwe Would you say this is ill-advised? I noticed you used 32 points as the cut-off in your own usage of this, but I figured if it's not outrageously expensive it might be a good thing to do for most manifolds.

src/spaces/jolt_physics_direct_space_state_3d.cpp
```diff
@@ -983,28 +1019,75 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
...
+#endif // JPH_DEBUG_RENDERER
+) const {
+	const JPH::PhysicsSystem& physics_system = space->get_physics_system();
+	const JPH::PhysicsSettings& physics_settings = physics_system.GetPhysicsSettings();
+
+	JPH::ManifoldBetweenTwoFaces(
+		p_hit.mContactPointOn1,
+		p_hit.mContactPointOn2,
+		p_hit.mPenetrationAxis,
+		physics_settings.mManifoldToleranceSq,
```
mihe, line 1074: @jrouwe Forgive me for pinging you twice, but should I also be adding the squared speculative contact distance here, like you do in your non-reducing collector?

src/spaces/jolt_physics_direct_space_state_3d.cpp
jrouwe: The 32 points is just for intermediate results. In PhysicsSystem::ProcessBodyPair it keeps accumulating contact points from different hits with the same normal until the buffer is full (64 is full, you get max 32 per hit so it prunes them if there are more than 32 in the buffer currently to ensure there's enough space). When all hits have been processed it does a final prune to limit it to 4:

https://github.com/jrouwe/JoltPhysics/blob/e60de1aa544dbdb77916b5ef410f2563e2b313bd/Jolt/Physics/PhysicsSystem.cpp#L1138-L1139

It's hard to see from this diff exactly what you are doing, but if you're only pruning once then 4 should be good. If you're combining many hits and run prune multiple times then you may want to do what I do to limit the amount of calls to prune.

mihe: Ah, gotcha. There's only one prune, so 4 it is then. Thank you!

yosoyfreeman:
Could this be part of the reason why Godot is able to get stable contacts with walls and Jolt does not?

mihe:
> Could this be part of the reason why Godot is able to get stable contacts with walls and Jolt does not?

That particular discrepancy is a result of me somewhat deliberately making the collision part of `body_test_motion`/`test_move` have a shorter reach than the depenetration/recovery part of it, which reduces ghost collisions by quite a bit, but causes things like `is_on_wall` to break, because you're simply always pushed away from any surface to the point where it can't be detected by the collision check.

Godot (from what I understand) does it the other way around, where the depenetration/recovery is ever so slightly shorter than the collision check instead, more or less guaranteeing that you'll be overlapping with something (although I've found this to be pretty inconsistent). I've experimented with doing it the same way, but no matter how I go about it I get an awful amount of ghost collisions.

I don't see any difference whatsoever from this change with the way the code is now, but it could very well be that it maybe has some effect with the way Godot does it. I'll take a look later.

It would be both awesome and frustrating if it actually did, considering how much time I've spent these past few weeks debugging this mess, yet never reflected over this (in hindsight) incredibly obvious discrepancy.

yosoyfreeman:
Thank you for looking into it!

Ghost collisions could be reduced in the built in move and slide by implementing a basic validation. It is impossible to collide with a plane that points in the direction of motion (It's surface is back to you).  That's how i'm reducing the impact of those on the sliding resolution. It will only slide along valid planes.

In my experience, unless you process any polygon individually to always prevent overlapping like it does in [this paper](https://www.peroxide.dk/papers/collision/collision.pdf ) it is impossible to create a perfect threshold that works for move and slide. There is always a possibility that you will end up inside a polygon, in which case move and collide can not move anymore and the character gets stuck. This happens as a result of the way move and collide works, independently of move and slide and block on walls. As long and move and collide can stop by a collision against a plane coplanar to the velocity, once you are inside a polygon there is no way out. So we would need to validate the collisions against the motion in move and slide and ignore planes that points towards the velocity(That is, you can not slide with the wall you are moving against). This would fix the case in which touching a wall and a floor at the same time causes block on wall to activate even if you are trying to leave the wall. We also need an un-stucking algortihm like [this](https://www.youtube.com/watch?v=AHtnuRkeP6Y)  to get out of the geometry once we reached an illegal position as a last resort.

So, at least in my experience and in my eyes(And i may be very much wrong), improvements on move and slide and needed in order for it to work properly. It is not just collision detection, move and slide needs a way of preventing this situation in the case they happen, because they will happen eventually and the code does not consider the situation of a full stuck, so there are not safeguards. The full implementation relies on collision being as perfect as a direct comparison of the shape against each poly in a range as shown in the paper, but that is not always the case for us. With iterative collision resolution we need to handle the case in which you ended stuck anyway.

src/spaces/jolt_physics_direct_space_state_3d.cpp
jrouwe: No you should not add that here, the parameter is named a bit strangely maybe (also this function was never intended to be used by anyone but me). 

image

In situation 1, the colliding faces are aligned so the clipped contact manifold (purple) should be returned. In situation 2, the colliding faces are not aligned and if the distance^2 between the left two points (orange) is bigger than `inSpeculativeContactDistanceSq` the point will be discarded and the contact manifold will only consist of the right point. In the query case you'd only specify a tolerance here, the way it is used in the physics engine is that the speculative contact distance is added to it because it should detect collisions that are further away (the bodies are not touching yet but we still want to know what the manifold will look like if the bodies move far enough in the time step to get into contact).

mihe: Great, thank you!

> also this function was never intended to be used by anyone but me

I got the sense that this might have been the case, judging by the debug rendering stuff that's in there. Would you rather see that some new (more formal/public) method gets added, or that I go about doing this in some other way, or do I go ahead with this implementation for now?

jrouwe: Go ahead for now, it means that it is slightly more likely that I will make a breaking change to your code

mihe: Will do. That's totally fine.

mihe:
> So, at least in my experience and in my eyes(And i may be very much wrong), improvements on move and slide are needed in order for it to work properly.

I appreciate the input. I must admit that despite having spent quite a bit of time debugging `move_and_slide` and its interactions with `body_test_motion` it's still somewhat of a black box to me, so what you're saying might be entirely correct, but I can't really say one way or the other. My general impression is that `move_and_slide` is in need of some changes though, like what you're saying, but beyond ripping out that `stop_all_motion` part of `floor_block_on_wall` (which may or may not be advisable) I don't have any concrete suggestions unfortunately.

Anyway, I've done some testing now with this PR using a `body_test_motion` implementation that more closely resembles the one Godot Physics uses, and I don't see any improvement, unfortunately. It still seems worthwhile to merge though, given that there is some sort of discrepancy here, so I'll go ahead with that.

PR #817, opened 2024-03-24
Fixed `body_get_direct_state` returning non-null with no space

### LINKED ISSUE #816: `JoltPhysicsServer` returns the direct state of bodies without a space in `_body_get_direct_state`
Hello everyone! I have encountered a bug in Godot Jolt: `JoltPhysicsServer3D::_body_get_direct_state` should return null if the body has previously being freed, as per the comment inside that function. However I believe it should also return null if the object's `space` is null. [Godot's implementation does this as well.](https://github.com/godotengine/godot/blob/99ff024f78f65ba0bc54fb409cfeca43ba2008fe/servers/physics_3d/godot_physics_server_3d.cpp#L933)

For some more context of how I discovered the bug and why it is relevant: When my player character interacts with a door the corresponding level scene gets loaded and added to the scene tree. The previous level gets removed from the scene tree, **but is not freed**. Instead I cache the level to save on load times in case the player returns through the door. The player stays inside the scene tree and gets teleported to the location of the door in the new level (while the scene tree is paused).
The problem I had with this is that every time I walked through a door, I consistently got the following error when calling `move_and_slide()` for the first time after the level transition has completed and the game got unpaused:

"ERROR: Failed to retrieve point velocity for 'StaticBody3D:<StaticBody3D#97676953793>'. Doing so without a physics space is not supported by Godot Jolt. If this relates to a node, try adding the node to a scene tree first."

The error message is referring to the floor of the previous level, which has been removed from the scene tree on purpose. I looked through the source code of Godot Jolt and of Godot and believe to have found the error. By adding the line

"QUIET_FAIL_NULL_D(body->get_space());"

below "QUIET_FAIL_NULL_D(body);" in `PhysicsDirectBodyState3D* JoltPhysicsServer3D::_body_get_direct_state(const RID& p_body)` the error disappears. I will be submitting a pull request with that one line and an adjustment to the comment above it shortly. Please let me know if you need more context or have suggestions for a better solution. Thank you!

tracefree:
Fixes #816 

This is my first ever contribution to an open source project, so please let me know if I'm doing anything wrong!

### DIFF
--- a/src/servers/jolt_physics_server_3d.cpp
+++ b/src/servers/jolt_physics_server_3d.cpp
@@ -1138,9 +1138,10 @@ PhysicsDirectBodyState3D* JoltPhysicsServer3D::_body_get_direct_state(const RID&
 	JoltBodyImpl3D* body = body_owner.get_or_null(p_body);
 
 	// Unlike most other server methods this one is meant to quietly return null if the body has
-	// since been freed, which is used in places like `move_and_slide` to determine whether a
-	// previously used platform has been freed or not.
+	// since been freed or removed from the scene tree, which is used in places like
+	// `move_and_slide` to determine whether a previously used platform has been freed or not.
 	QUIET_FAIL_NULL_D(body);
+	QUIET_FAIL_NULL_D(body->get_space());
 
 	return body->get_direct_state();
 }


tracefree:
I should mention that I am still relatively new to C++ and don't really know what the performance implications of my fix are. I copied the way the condition in the line above mine gets checked, [but Godot itself does the same check differently.](https://github.com/godotengine/godot/blob/99ff024f78f65ba0bc54fb409cfeca43ba2008fe/servers/physics_3d/godot_physics_server_3d.cpp#L933)

tracefree:
Thank you for the quick merge, and for your amazing work on bringing good physics to Godot!

PR #830, opened 2024-04-12
Changed the way soft body edges are created

jrouwe:
- The pinned vertices weren't initialized before calling Optimize, this meant that the edge sorting algorithm could not determine an optimal order.
- Instead of manually creating the edges, now using the CreateConstraints function which will also create shear edges which results in a better cloth simulation.
- Limited the minimal compliance to 1.0e-5 instead of 1.0e-6 to make the cloth more stable.


jrouwe:
Based on discussion in jrouwe/JoltPhysics#1042.

The main issue was that the pinned vertices weren't set before calling Optimize(), after doing this the scene became stable again.

Because it is on the border of being unstable (you can see small jitters at the top of the vertical hanging cloth) and all the Jolt examples use a compliance of 1.0e-5, I've changed the min value to match that.

Finally, I changed the way edges are created because the constraints were created like this:

image

where this configuration is symmetric and more stable:

image

mihe:
Awesome, thank you so much. I threw in some minor additional changes.

I also changed the title since there hasn't technically been a regression, as I haven't pulled in jrouwe/JoltPhysics#1042 yet.

PR #836, opened 2024-04-22
Simplified `JoltBodyImpl3D::get_principal_inertia_axes`

jrouwe:
I was going through the `TODO(mihe)` tags and noticed this one. The calculation that's there currently is quite expensive (`GetMassProperties` does quite a lot of things) and I think the calculation is quite trivial since Jolt also splits the inertia in a diagonal matrix and a rotation matrix.

Note: Code is totally untested as I don't know how to test it!

### DIFF
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -301,13 +301,10 @@ Basis JoltBodyImpl3D::get_principal_inertia_axes() const {
 	const JoltReadableBody3D body = space->read_body(jolt_id);
 	ERR_FAIL_COND_D(body.is_invalid());
 
-	// TODO(mihe): See if there's some way of getting this directly from Jolt
-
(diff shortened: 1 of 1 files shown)


mihe:
Oh yeah, I've been meaning to ask you about that one.

I couldn't at the time quite see what that whole circus with Godot's `Basis::diagonalize` would map to with Jolt, but I suppose that is in fact just a roundabout way of getting the inertia rotation, which you of course have readily available in Jolt.

I'll give it a try tomorrow!

(You also prompted me to check out my TODO comments, and I'm realizing now I really need to implement face collection for the ray shape, given that `InternalEdgeRemovingCollector` depends on it, but it should just be about using `GetSupportingFace` I guess?)

jrouwe:
Yes,

```
// TODO(mihe): Implement face collecting
```

should be something like:

```
p_shape2->GetSupportingFace(hit.mSubShapeID2, ray_direction2, p_scale2, p_center_of_mass_transform2, result.mShape2Face);
```

(and you'll need to remove the `const` from `result`).

PR #837, opened 2024-04-24
Added face collection to custom ray shape

mihe:
This adds support for `JPH::ECollectFacesMode::CollectFaces` to `JoltCustomRayShape`, which is what powers `SeparationRayShape3D`.

This should hopefully make it play nicer with things like `JPH::InternalEdgeRemovingCollector` that makes use of these faces in its algorithm.


jrouwe:
The other shapes never return a face with only 1 vertex, all algorithms that deal with faces will fall back to the contact point when an empty face is returned.

PR #842, opened 2024-05-04
Implementing new JPH::Reallocate function

jrouwe:
This won't compile at the moment, but if you upgrade to the latest version of Jolt you'll need to implement this function. Everything else still seems to compile and work as it is supposed to.


mihe:
Much appreciated!

mihe:
Superseded by #848.

PR #860, opened 2024-05-18
Changed `body_test_motion` to discard by angle rather than depth

### LINKED ISSUE #626: `move_and_slide` is slower than with Godot Physics
It appears that version 0.7.0 (most likely #559) introduced a performance regression to `move_and_slide`, causing it to take roughly twice as long as before, leading to it now being slower when compared to Godot Physics.

Here's how to reproduce it:

1. Download this project: [move_and_slide_regression.zip](https://github.com/godot-jolt/godot-jolt/files/12777042/move_and_slide_regression.zip)
2. Download debug symbols (if you're going to profile): [godot-jolt_v0.8.0-stable_symbols.zip](https://github.com/godot-jolt/godot-jolt/releases/download/v0.8.0-stable/godot-jolt_v0.8.0-stable_symbols.zip)
3. [Build Godot from source](https://docs.godotengine.org/en/stable/contributing/development/compiling/index.html) (if you're going to profile)
4. Run the main scene
5. Note the average delta displayed on the screen
6. Switch physics engine to Jolt
7. Run the main scene again
8. Note the average delta being higher

When picking out some average-looking physics ticks in Superluminal, the total time spent in `move_and_slide` in this particular scene is as follows:

Godot Physics (godotengine/godot@58f0cae): 4.74 ms
Godot Jolt ([0.6.0-stable](https://github.com/godot-jolt/godot-jolt/releases/tag/v0.6.0-stable)): 3.41 ms
Godot Jolt ([0.7.0-stable](https://github.com/godot-jolt/godot-jolt/releases/tag/v0.7.0-stable)): 5.70 ms
Godot Jolt ([0.8.0-stable](https://github.com/godot-jolt/godot-jolt/releases/tag/v0.8.0-stable)): 6.11 ms

(Note that these fluctuate quite a bit, so the difference between 0.7.0 and 0.8.0 seems to be within margin of error and probably not the result of further regression.)

When looking at `move_and_slide` more closely, it appears that Godot Jolt is returning collisions from `body_test_motion` that causes Godot to believe that some of them are not contacting the floor, causing it to run `apply_floor_snap` [here](https://github.com/godotengine/godot/blob/0ca8542329888e8dccba89d59d3b728090c29991/scene/3d/physics_body_3d.cpp#L1609), which ends up taking as much time as the slide iterations themselves.

### LINKED ISSUE #643: `move_and_slide` jitters when colliding with flat wall and angled surface
When running against a gap that is barely not large enough for the collision shape the CharacterBody3D will jitter and jumping is affected.

**Cylinder Shape**: The CharacterBody3D will jitter and is_on_floor() will flip between true and false. Jumping is nearly impossible since you do not get the right amount of height from a jump.

**Capsule Shape**: The CharacterBody3D will jitter and is_on_floor() will flip between true and false. Jumping is nearly impossible since you do not get the right amount of height from a jump.

**Box Shape**: The CharacterBody3D will jitter **but is_on_floor() is right**. Jumping is nearly impossible since you do not get the right amount of height from a jump.

_The jumping bug does still happens when the gap is much smaller than the width of the collision shape._

Setting the amount of max_slides to a value much higher ( I use ~20) than default will fix the jumping issue, but the jitter issue and is_on_floor() issue is still present. This happens with 60 physics ticks per second and 100 physics ticks per second with no change. These scenarios occur with default jolt settings on 0.9.0 and 0.8.0. Not entirely sure if this is a known issue or not. Since it affects the most common collision shape I thought it would be worth making this issue anyways.

MRP is included.  The .godot and addons folder are not present in the MRP so godot-jolt needs to be installed.
Setting the max amount of slides is commented out in the _ready() function. is_on_floor() is printed at the beginning of every physics tick in the console. All that is needed to change collision shapes is to change CShape in player.tscn (The default sizes of the named collision shapes are what should be used).

[MRP-Jittering.zip](https://github.com/godot-jolt/godot-jolt/files/12890919/MRP-Jittering.zip)

mihe:
Fixes #626.
Fixes #643.

This applies the change to `PhysicsServer3D.body_test_motion` that was suggested by @yosoyfreeman in #812, which effectively reverts the hack introduced by #559 and replaces its depth-based contact discard with an angle-based one instead.

Note that this also impacts `PhysicsBody3D.move_and_collide` as well as `PhysicsBody3D.test_move` and of course `CharacterBody3D.move_and_slide`.

This will possibly lead to a regression in ghost collisions for some scenarios, like the one seen in #529, but the downsides of #559 have proved to be too great to keep it around.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Breaking changes are denoted with ⚠️.
 
 ### Changed
 
+- ⚠️ Changed the way that the `body_test_motion` method of `PhysicsServer3D` discards contacts,
+  which should fix issues related to jitter/ping-ponging. Note that this can also result in *more*
+  ghost collisions under certain conditions. This affects `move_and_slide`, `move_and_collide` and
+  `test_move`.
 - ⚠️ Changed the inertia of shapeless bodies to be `(1, 1, 1)`, to match Godot Physics.
 - Changed `SeparationRayShape3D` to not treat other convex shapes as solid, meaning it will now only
   ever collide with the hull of other convex shapes, which better matches Godot Physics.
--- a/src/spaces/jolt_physics_direct_space_state_3d.cpp
+++ b/src/spaces/jolt_physics_direct_space_state_3d.cpp
@@ -752,7 +752,7 @@ bool JoltPhysicsDirectSpaceState3D::test_body_motion(
 		collided = _body_motion_collide(
 			p_body,
 			transform.translated(p_motion * unsafe_fraction),
-			(float)p_motion.length(),
+			p_motion,
 			p_margin,
 			p_max_collisions,
 			p_result
@@ -1108,11 +1108,15 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_cast(
 bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 	const JoltBodyImpl3D& p_body,
 	const Transform3D& p_transform,
-	float p_distance,
+	const Vector3& p_motion,
 	float p_margin,
 	int32_t p_max_collisions,
 	PhysicsServer3DExtensionMotionResult* p_result
 ) const {
+	if (p_max_collisions == 0) {
+		return false;
+	}
+
 	const JPH::Shape* jolt_shape = p_body.get_jolt_shape();
 
 	const Vector3 com_scaled = to_godot(jolt_shape->GetCenterOfMass());
@@ -1147,18 +1151,26 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 		return collided;
 	}
 
-	// HACK(mihe): Without this minimum contact depth we can sometimes end up with very shallow
-	// contacts that end up affecting the outcome of things like `floor_block_on_wall`, where after
-	// one of the recovery iterations (in Godot, not here) we can still find ourselves penetrating a
-	// wall ever so slightly, which `move_and_slide` will interpret as being trapped in a corner and
-	// stop the character altogether. We still need distances smaller than this (like none at all)
-	// to actually emit contacts though, so we clamp it by the distance moved.
-	const float min_contact_depth = MIN(0.0001f, p_distance);
-
 	int32_t count = 0;
 
 	for (int32_t i = 0; i < collector.get_hit_count(); ++i) {
 		const JPH::CollideShapeResult& hit = collector.get_hit(i);
+
+		const float penetration_depth = hit.mPenetrationDepth + p_margin;
+
+		if (penetration_depth <= 0.0f) {
+			continue;
+		}
+
+		const Vector3 normal = to_godot(-hit.mPenetrationAxis.Normalized());
+
+		if (p_motion.length_squared() > 0) {
+			const Vector3 direction = p_motion.normalized();
+
+			if (direction.dot(normal) >= -CMP_EPSILON) {
+				continue;
+			}
+		}
 
 		JPH::ContactPoints contact_points1;
 		JPH::ContactPoints contact_points2;
@@ -1177,12 +1189,6 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 			contact_points2.push_back(hit.mContactPointOn2);
 		}
 
-		const float penetration_depth = hit.mPenetrationDepth + p_margin - min_contact_depth;
-
-		if (penetration_depth <= 0.0f) {
-			continue;
-		}
-
 		const JoltReadableBody3D collider_jolt_body = space->read_body(hit.mBodyID2);
 		const JoltShapedObjectImpl3D* collider = collider_jolt_body.as_shaped();
 		ERR_FAIL_NULL_D(collider);
@@ -1199,7 +1205,7 @@ bool JoltPhysicsDirectSpaceState3D::_body_motion_collide(
 			PhysicsServer3DExtensionMotionCollision& collision = p_result->collisions[count++];
 
 			collision.position = position;
-			collision.normal = to_godot(-hit.mPenetrationAxis.Normalized());
+			collision.normal = normal;
 			collision.collider_velocity = collider->get_velocity_at_position(position);
 			collision.collider_angular_velocity = collider->get_angular_velocity();
 			collision.depth = penetration_depth;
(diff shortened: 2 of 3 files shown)


yosoyfreeman:
Thank you a lot for your work Mihe!

mihe:
It turned out that this did not in fact fix #626, since `apply_floor_snap` still runs when walking across a perfectly parallel floor, which I probably should have caught before pushing this. It's not really an argument for reverting this though, since it was arguably worse before, but it's still somewhat unfortunate.

However, I am also slightly worried over the fact that this change results in zero slide collisions when a character is walking over a perfectly parallel floor like that. That's obviously to be expected given that this change discards any such planes, but I've seen a couple of examples of cases where people rely on these slide collisions to gather information about the surface they're walking on, which will now be more difficult.

It would have been nice if `move_and_slide` stored the wall/floor contacts explicitly rather than force people to rely on the slide collisions.

yosoyfreeman:
Hi Mihe!

> It turned out that this did not in fact fix https://github.com/godot-jolt/godot-jolt/issues/626, since apply_floor_snap still runs when walking across a perfectly parallel floor, which I probably should have caught before pushing this. It's not really an argument for reverting this though, since it was arguably worse before, but it's still somewhat unfortunate.

It is unfortunate indeed, but to be expected. Move and slide is not capable of knowing is touching a floor unless you directly collide against one. This means that, even if it works perfectly, a call to apply_floor_snap **will** be done, as is the only way the system has to retrieve floor info. Its the same problem that appears while moving parallel to a wall, it will report nothing.

> However, I am slightly worried over the fact that this change results in zero slide collisions when a character is walking over a perfectly parallel floor like that. That's obviously to be expected given that this change discards any such planes, and I would probably argue that such a movement shouldn't result in any slide collisions either, but I've seen a couple of examples where people rely on these slide collisions to gather information about the surface they're walking on, which will now be more difficult.

To fix this issue with walls, once i touched one i add a minimal "force" or velocity towards it. This is enough to make the character collide every frame with it without altering user input too much, so i can gather the data without skips.

Could this work in this scenarios? I know is a little of a hack, but seems less of a hack than what move and slide was actually doing. Maybe a small gravity when on floor would be enough for this cases.

> It would have been nice if move_and_slide stored the wall/floor contacts explicitly, rather than force people to rely on the slide collisions (where they typically check the normals themselves), in which case apply_floor_snap could have stored its collision results there instead of them being discarded completely.

The problem is that move and slide does not have any way of knowing nothing that is not the first contact on the direction of the movement or apply floor snap. What is reported by is on wall or is on floor is basically what is reported in one of the slide collisions as the last floor or wall hit. They are just a quicker way of looking for it. I learning my way to c++, so i could start trying to implement some hacks... But they would be that, hacks.  Because of course we would want to rewrite it as soon as we would have better sweeping, normal data etc.

mihe:
> It is unfortunate indeed, but to be expected. Move and slide is not capable of knowing is touching a floor unless you directly collide against one. This means that, even if it works perfectly, a call to apply_floor_snap **will** be done, as is the only way the system has to retrieve floor info. Its the same problem that appears while moving parallel to a wall, it will report nothing.

This is however not how Godot Physics behaves, which seemingly never runs `apply_floor_snap` when just moving along a level surface like that, which bothers me quite a bit. I've spent a lot of time comparing the two implementations trying to figure out why that is, but have not been able to track down the exact cause.

I've always assumed it has something to do with the type of minimum contact depths that Godot Physics uses, where it will deliberately undershoot the recovery step, but if I try to replicate the same thing on my end I get a prohibitive amount of ghost collisions, to the point where it's almost impossible to walk up an inclined surface, hence why I took the opposite approach in #559 where I undershoot the final collision check step instead.

> Could this work in this scenarios? I know is a little of a hack, but seems less of a hack than what move and slide was actually doing. Maybe a small gravity when on floor would be enough for this cases.

Anything like this would presumably need to happen in `move_and_slide` itself. `body_test_motion` is just a small piece of it all, and has no idea about any sort of orientation of the thing being moved. In fact, since it powers `move_and_collide` as well it doesn't even necessarily have to involve a `CharacterBody3D`.

> The problem is that move and slide does not have any way of knowing nothing that is not the first contact on the direction of the movement or apply floor snap. What is reported by is on wall or is on floor is basically what is reported in one of the slide collisions as the last floor or wall hit.

Yes, I know, but what I'm saying is that people currently rely on slide collisions in a way where they iterate through them all, check the normal of them, and then decide based on the angle of that normal whether that collision is a wall or a floor, and then use those accordingly. This is obviously a bit awkward/redundant since `move_and_slide` has already done the work (in `_set_collision_direction`) to decide which slide collisions map to what surface.

This is made worse by the fact that `apply_floor_snap` does not actually store its collision results as additional slide collisions, which is why we end up with zero slide collisions when walking along a level surface in Godot Jolt.

But if `move_and_slide` actually stored its collisions as `floor_collisions` and `wall_collisions` or something along those lines, then there's presumably no real reason why `apply_floor_snap` couldn't just add its collision results to `floor_collisions` and people could rely on those instead of inferring it from the slide collisions.

That obviously does nothing to fix the issue of wall collisions, but it would be an improvement at least.

yosoyfreeman:
Hi Mihe, i got some time free today and checked the code with what you told me in mind.

I agree that adding floor snap results as slide collision can be a solution. I could _try_ to do that (Still learning c++ and godot codebase). However, i do think the reason they are not stored is because they are technically not slide collisions, in the sense that they do not trigger the sliding behavior or alter the velocity in any way. However, the responsibilities of checking the environment and reacting to it are already mixed. Since no breaking changes can be done by now, this would allow to at least provide more information to the user. It does not require additional parameters or API either, unlike adding a new floor, wall and ceiling categories. As i see it, calling a floor collision an slide collision is a minor hack compared to some of the other ones that are already there.

If you think this could improve things, i can look into it and try to do a PR. As i said, is probably the only one with some chance to get accepted right now and no breaking changes are required.

mihe:
> As i said, is probably the only one with some chance to get accepted right now and no breaking changes are required.

What do you mean by "get accepted right now"? As in Godot 4.3-beta1 or Godot 4 in general?

I would perhaps argue that adding more collisions to the list of slide collisions (that may or may not belong there) is an actual breaking change (albeit a small one) whereas adding new categories is not. I have a hard time imagining either one making it into Godot 4.3 though, given that we've entered the beta phase already, and Godot Physics doesn't seem to need the change.

If you're just talking about Godot 4 in general, I wouldn't shy away from adding to the API necessarily. It seems to me like adding those new lists/categories would be fairly simple to provide a reason for, seeing as how users work around it by checking the normal when using either physics engine.

I am however realizing now that even if these separate categories were to be added, we would have a hard time categorizing `get_last_slide_collision()`, since that particular collision (which is probably important to some people) may or may not be in any of these new lists. Adding some kind of `is_floor()` and `is_wall()` to `KinematicCollision3D` would have been a nice option instead, but unfortunately it's shared with `move_and_collide`, which has no notion of walls and floors, so that doesn't seem right.

Issue #855, opened 2024-05-15
JoltGeneric6DOFJoint3D angular limit gizmo not correct


tlobig:
while playing around with this addon, which I really like, I figured there is a small issue with the Jolt-version of the 6DOF joints.

The short version: The angular limit gizmos only show the actual limits correctly if the joint itself is rotated by 90 degrees.
![screenshot in editor](https://github.com/godot-jolt/godot-jolt/assets/10432347/551a1756-446d-4911-bf79-fc77e3ba8333)

![screenshot when running the scene, notice middle joint is not limiting angle as previewed with Gizmo](https://github.com/godot-jolt/godot-jolt/assets/10432347/fbf58aa4-3e14-47b7-a429-ee50fff95806)

steps to reproduce:
should be obvious, have a static body  and a rigid body connected by a Jolt 6DOF joint, set an angular limit that's relatively narrow to see the effect AND the gizmo, add a camera and run the scene  to see the angular limit is not solved as previewed by the gizmo.

minimal project to reproduce:
[jolt_gizmo_minimal.zip](https://github.com/godot-jolt/godot-jolt/files/15316517/jolt_gizmo_minimal.zip)

Godot-Jolt version: 0.12.
Godot version: 4.2.1 stable

mihe:
Thank you for reporting this. This has been in the back of my mind for a while now, but I never got around to it.

I've spent a silly amount of time trying to make sense of these joint gizmos, ever since I first implemented them, which I realize in hindsight was largely a result of me getting tripped up by the way that single-body joints work in Godot Physics, which I for some reason felt the need to accommodate for (over two-body joints). 

With this behavior having changed in #725 I should rightfully have changed the gizmo accordingly, but forgot to.

Anyway, this should be fixed when #862 is merged. It now seems to line up nicely with Jolt's own debug visualization, as well as Nvidia Omniverse, which happens to use a very similar visualization in its USD Composer editor.

tlobig:
extremely cool and swift execution. nicely done.

tlobig:
just compiled the master - is this the intended display of the gizmos?
![grafik](https://github.com/godot-jolt/godot-jolt/assets/10432347/21ddedf0-9b1f-4be0-b0fb-bbe212dffa32)
![grafik](https://github.com/godot-jolt/godot-jolt/assets/10432347/1ee0c664-ab3b-42e9-8500-e58d7df8dc74)

[...]

PR #872, opened 2024-05-26
fix to axis dependent mismatch of jolt joint angular limit gizmo

tlobig:
this is my supposed fix to the jolt joing angular gizmos being wrong.

My initial idea worked only for the x axis, the other axes need to be treated differently.

You can just take this PR as an inspiration and create a cleaner solution, I wont mind. Or edit this one to suit your needs.


mihe:
Thank you for the attempt! I'll see what I end up doing about this.

(See my comment in #855 for more details.)

Issue #477, opened 2023-07-03
Add support for web


TranquilMarmot:
Similar to https://github.com/godot-jolt/godot-jolt/issues/354 and https://github.com/godot-jolt/godot-jolt/issues/353, I wonder if it would be possible to support jolt on the web (HTML5 export)?

mihe:
Thank you for the feature request.

Jolt does technically support web builds through [JoltPhysics.js](https://github.com/jrouwe/JoltPhysics.js), but I haven't really explored what it would look like to make use of it from Godot Jolt, nor what loading extensions looks like for Godot's web exports.

I must admit that I'm not super keen on supporting this, which is why I didn't create an issue for it. Although the situation is certainly much better these days, browsers tend to have frustrating discrepancies that I'd rather not spend my time chasing down. Throwing Emscripten and WebAssembly into the mix will probably bring its own set of challenges as well.

I'll leave this open to better gauge what the demand is for this, but for the time being this will not be on the roadmap.

However, if someone else with more experience with Emscripten and WebAssembly wants to do some exploratory work, to give me a better sense of how it all would work and what changes would be necessary, then I'd be willing to allocate some time for accommodating that.

TranquilMarmot:
Yeah, even reading [Godot's docs on exporting for the web](https://docs.godotengine.org/en/stable/tutorials/export/exporting_for_web.html) there are so many warnings 😅 It is definitely a somewhat painful thing to support right now.

jrouwe:
I concur that supporting this is painful. I managed to make it work for Jolt, but the code, compile, debug loop is extremely slow. The debugger is almost unusable. I had to resort to printf debugging in many cases and even then it manages to not show the printfs. There are indeed bugs in browsers, e.g. https://github.com/jrouwe/JoltPhysics/issues/577 where even the HelloWorld application doesn't work on macOS Safari.

bikemurt:
Just chiming in here - not sure if there's been any more decisions regarding Jolt Physics and the web. One of the most frustrating things I'm dealing with right now is very basic collision shapes "jittering" unnecessarily with the vanilla physics engine. I'm using very basic shapes such as boxes and cylinders, and no matter how I tweak all the parameters, the physics engine just seems to do it wrong.

I've tested in Jolt Physics and it works with no errors whatsoever, but of course, I am developing for the web, hence my post...

[...]

PR #902, opened 2024-06-18
Added Initial (experimental) web support.

tetious:
This adds web support, tested with a medium sized project on:

- MacOs (Firefox and Chrome) :green_circle: 
- Linux (Firefox and Chrome) :green_circle: 
- ChromeOS (Multiple low-ram Chromebooks) :yellow_circle:  (see TODO below)

# Please try it out!

## Option A

1. Grab the extension files [here](https://orcling.com/godot-jolt/godot-jolt-web-5497271.zip) and unzip them over your `addons/godot-jolt` folder.

## Option B

You can build everything yourself following the build instructions in this PR (see building.md). 

## Please report back!

Either option you pick, please let us know how it works for your project.

# TODO

- [x] Test with Emscripten 3.1.62 when it is released (both multi and single-threaded, and against the "official" export templates)
- [ ] Investigate warnings in godot-cpp when LTO is enabled:
```
wasm-ld: warning: function signature mismatch: _ZNK5godot11AudioStream21_instantiate_playbackEv
>>> defined as () -> void in External/Build/godot-cpp/EditorDistribution/libgodot-cpp.a(unity_3_cxx.cxx.o)
>>> defined as (i32, i32) -> void in lto.tmp

wasm-ld: warning: function signature mismatch: _ZNK5godot14TextureLayered15_get_layer_dataEi
>>> defined as () -> void in External/Build/godot-cpp/EditorDistribution/libgodot-cpp.a(unity_23_cxx.cxx.o)
>>> defined as (i32, i32, i32) -> void in lto.tmp
```
- [x] Investigate crash on low-memory Chromebook:
I've now isolated this to impacting any trimesh based collider. If any body (or ray) interacts with a body that has a trimesh collider, it immediately crashes with a memory access violation.
I'm not sure yet why this only happens in ChromeOS, but I don't think it has anything to do with memory pressure, as I've been able to repro on a device with more RAM.
- [ ] Investigate repeatable crash with trimesh colliders in ChromeOS.
- [ ] Figure out how to support threaded builds without passing a enscripten global build flag
- [x] CI setup
- [x] Basic smoke testing on more platforms/browsers


mihe:
I appreciate the work you've put in here, but just so there's no misunderstanding, my stance on supporting this hasn't changed much since my original reply in #477. If anything, I have much less time to dedicate to this extension these days, so I have even less bandwidth to spare for this, especially when considering whatever bug reports or CI issues that might arise from this down the road. It already costs me a full day to cut a release for this extension, due to smoke testing all the various platforms and architectures and their respective deployment methods. I'm not super excited about adding more permutations to that list.

I'm not averse to merging this (when it's done) but there needs to be some sort of commitment from somebody with regards to taking on the above mentioned extra work.

If everything goes as planned that commitment shouldn't have to extend much further than Q1-ish 2025, at which point this extension will ideally have been merged into Godot as an official module and released as part of Godot 4.4, shortly after which I will likely archive this repository. Don't quote me on that timeline though. I'm sure something will delay it.

tetious:
Totally understandable. I really appreciate all the work you've put into this extension, especially all the documentation and infrastructure. It was a pleasure to work on, in no small part to the hacking document and the overall structure and build process.

> needs to be some sort of commitment from somebody with regards to taking on the above mentioned extra work.

I'm happy to commit to this. I'm working on a project that relies heavily on Joints, and have found GodotPhysics sufficiently unpredictable and incomplete that I need another option. So I need to see this through one way or another, and I'd love it to be in a form that can benefit others.

Here is what I propose: Emscripten 3.1.62 will likely drop soon, and Godot 4.3 final too. I'm hoping the combination of those two  (or a followup 4.3.x) will eliminate the need to custom build the export templates in the medium turn, but in the mean time I can produce those semi-automatically and host them.

With the final release of 4.3 (or soon thereafter), we can add web support under an unsupported "experimental" tag to gather feedback, etc. I will handle the web-related bug reports (though I will be of limited help with any physics/math stuff, alas) and CI setup for this and any subsequent releases until Godot 4.x takes it in. 

I can start getting the various CI stuff going now in a separate MR?

Thanks again for your help getting this going and for your stewardship of this extension!

mihe:
> I'm hoping the combination of those two (or a followup 4.3.x) will eliminate the need to custom build the export templates in the medium term, but in the mean time I can produce those semi-automatically and host them.

I guess we'll have to see how these things develop, but I'm not super keen on relying on third-party export templates. I would probably prefer to see that this PR remains unmerged until Godot's official export templates are fixed.

> I can start getting the various CI stuff going now in a separate MR?

Any reason why you couldn't do it in this PR?

tetious:
> I would probably prefer to see that this PR remains unmerged until Godot's official export templates are fixed.

Totally fine with me. I'm hopeful this will happen with the 4.3 release.

> Any reason why you couldn't do it in this PR?

Nope; happy to. Wasn't sure you'd want both things in one. :)

mihe:
> Nope; happy to. Wasn't sure you'd want both things in one. :)

Not a problem at all.

I'll turn on the CI stuff for you in this PR. Note that there are quite a few CI jobs, due to the combinatory explosion of platforms and architectures, so it takes roughly an hour for each CI run to complete fully, and much longer than that if there are others queued up obviously.

mihe:
Oh, and I don't run the packaging job as part of the regular CI, so you'll have to run that manually on your own fork to test that, which you should be able to do by pressing the "Run workflow" button [here](https://github.com/tetious/godot-jolt/actions/workflows/package.yml).

mihe:
Also, I realize this deviates from the Godot conventions, but for what could end up being a long-running PR like this one, I would appreciate if you could refrain from force-pushing and instead commit/merge like normal. It makes incremental reviews easier.

.github/workflows/build-web.yml
```diff
@@ -0,0 +1,65 @@
...
+        config: [debug] #, development, distribution]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup emsdk
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: 3.1.39
+          actions-cache-folder: 'emsdk-cache'
```
mihe, line 50: Looking at [Godot's usage](https://github.com/godotengine/godot/blob/0a83e7c5dac550afb26e26684cbb6e4d6c139f5e/.github/workflows/web_builds.yml#L44-L49) of this GitHub Action I'm seeing them provide a custom `cache-key` as opposed to using [the default](https://github.com/mymindstorm/setup-emsdk/blob/6ab9eb1bda2574c4ddb79809fc9247783eaf9021/src/main.ts#L30) one, which to me looks like it essentially prevents caching inbetween different pull requests, since it includes the `github.ref` in it. I wonder why that is.

CMakeLists.txt
```diff
@@ -41,6 +41,12 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL Linux)
 	list(APPEND sources ${info_file})
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
+	# This is a hack to workaround the lack of support for shared libraries in the toolchain file shipped with Emscripten
+	# Once this PR (or similar) lands (https://github.com/emscripten-core/emscripten/pull/16281) we can remove this
+	set_property(GLOBAL PROPERTY TARGET_SUPPORTS_SHARED_LIBS TRUE)
+endif()
```
mihe, line 48: I would put this in either `GodotJoltPrelude.cmake` or `GodotJoltSetup.cmake`, which run before and after the `project(...)` call respectively (and which probably should be included with `CMAKE_PROJECT_INCLUDE[_BEFORE]` instead). I'm not sure whether `CMAKE_SYSTEM_NAME` will be set before `project(...)`, so you might want to put it somewhere in `GodotJoltSetup.cmake` instead.

CMakeLists.txt
```diff
@@ -79,6 +86,11 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL iOS)
 	set(suffix_platform _ios)
 elseif(CMAKE_SYSTEM_NAME STREQUAL Android)
 	set(suffix_platform _android)
+elseif(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
+	set(CMAKE_SHARED_LIBRARY_CREATE_C_FLAGS "-sSIDE_MODULE=1")
+	set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "-sSIDE_MODULE=1")
```
mihe, line 91: I'm not seeing anything about these in the reference documentation. What's the purpose of this exactly?

CMakeLists.txt
```diff
@@ -79,6 +86,11 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL iOS)
...
+	set(CMAKE_SHARED_LIBRARY_SUFFIX ".wasm")
```
mihe, line 92: I would have thought CMake would do this automatically? What's the default suffix?

CMakeLists.txt
```diff
@@ -327,7 +342,11 @@ else() # ANDROID/UNIX
 	endif()
 
 	set(so_file $<TARGET_FILE:godot-jolt>)
-	set(debug_file ${so_file}.debug)
+	if(EMSCRIPTEN)
```
mihe, line 345: Does it maybe make more sense to just have Emscripten have its own `elseif(EMSCRIPTEN)` in the top-level if-statement here? It seems like there are enough subtle differences to warrant it.

CMakePresets.json
```diff
@@ -249,6 +249,18 @@
         "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/cmake/toolchains/android/x86.cmake"
       }
     },
+    {
+      "name": "emscripten-base",
+      "inherits": ["linux-base"],
+      "hidden": true,
+      "environment": {
+        "CXXFLAGS": "$penv{CXXFLAGS} -msimd128 -msse4.2 -mavx",
+        "LDFLAGS": "$penv{LDFLAGS} -msimd128 -msse4.2 -mavx -Wno-limited-postlink-optimizations -sALLOW_MEMORY_GROWTH=1 -sALLOW_TABLE_GROWTH=1"
```
mihe, line 258: What's the reasoning behind enabling SSE 4.2 and AVX? Godot targets SSE2.

CMakePresets.json
```diff
@@ -363,6 +375,12 @@
       "name": "macos-android-x86",
       "displayName": "Android, x86",
       "inherits": ["macos-base", "android-base", "android-x86-base"]
+    },
+    {
+      "name": "web-wasm32",
+      "inherits": ["emscripten-base","dev-base"],
```
mihe, line 381: Missing a space after the comma here.

CMakePresets.json
```diff
@@ -363,6 +375,12 @@
...
+      "displayName": "Web wasm32",
```
mihe, line 382: Add a comma after "Web" here.

mihe (changes requested):

.github/workflows/build-web.yml
tetious: Huh, interesting. Looks like they were having problems with cache contention with multiple concurrent builds. https://github.com/mymindstorm/setup-emsdk/issues/20#issuecomment-1898170886 

I do think maybe the branch name would have been sufficient, but I don't know too much about what and when they build.

CMakeLists.txt
tetious: Yeah, I didn't either. My understanding is that it allows you to override the flag that instructs the compiler to produce a dynamically linked/shared library.

I nabbed this from here: https://github.com/emscripten-core/emscripten/pull/16281/files

Emscripten's support for dlink is still new and a bit shaky, so they don't have support rolled into their toolchain yet. This can probably all go away once that PR lands. (though it seems like it has stalled, so...)

I'll add a comment to clarify.

CMakeLists.txt
tetious: The default is .so and chromium was refusing to load them until I renamed to .wasm. (Godot does this too.)

CMakeLists.txt
tetious: I almost did it that way but got lazy. Your nudge confirms it was the correct choice. :grinning: Will fix!

CMakePresets.json
tetious: Not much reasoning, haha. I'm a bit out of my depth in certain areas of this. Learning as I go! My thought was that targeting the "best" option supported by emscripten would be optimal, since they'd need to be downgraded by the VM if unsupported by the platform anyway. 

Pretty sure I nabbed this from the Jolt js port? 

Anyway, I can change this to mirror Godot.

tetious: Hmm. It seems if I don't either disable SSE completely or pass -msse4, jolt (the upstream library, not this plugin) fails to compile.

For now, I'll remove the avx flag and keep sse4.

.github/workflows/build-web.yml
mihe: Right, well, that won't be a concern for this repository I guess, so might as well use the default key.

CMakeLists.txt
mihe: Judging by how I've seen other `*_CREATE_*` flags used in CMake, I'm assuming this is just used by toolchains to add to the default set of `C[XX]_FLAGS`, so I would imagine this `-sSIDE_MODULE=1` could also just be added as part of the `target_compile_options(...)`.

tetious: Likely so, though in this case I personally think it makes more sense to keep it isolated, as it really isn't a compile option that is meant to be changed. It would normally be buried away with other low(er) level platform-specific implementation details, and _hopefully_ won't be needed long-term.

It's kinda like a non-user serviceable part in a "fancy" IoT enabled coffee maker. 😼

mihe: Are you sure this is actually doing anything to begin with? I tried setting it myself just now, both before and after `project(...)` as well as in the toolchain file for Android, and none of them seemed to result in this flag actually being added to any of the Ninja files.

tetious: Yep! It warns at the final link:
```
em++: warning: linking a library with `-shared` will emit a static object file.  This is a form of emulation to support existing build systems.  If you want to build a runtime shared library use the SIDE_MODULE setting. [-Wemcc]
```

And then this when I try to run:
```
need the dylink section to be first
displayFailureNotice @ tmp_js_export.html:140
Promise.then (async)
(anonymous) @ tmp_js_export.html:186
(anonymous) @ tmp_js_export.html:190
```

The resulting output is ~55 MiB vs ~4.5 MiB too, which I didn't really expect. :shrug:

mihe: Fair enough.

Either way, I'm not all that convinced about using these `*_CREATE_*` variables. This `-sSIDE_MODULE` flag is documented quite clearly [here](https://emscripten.org/docs/compiling/Dynamic-Linking.html#load-time-dynamic-linking) so I don't see much of a point trying to isolate it. None of the flags added through `target_compile_options(...)` are "user-serviceable" in the first place, as far as I know, unlike modifications to the global `C[XX]_FLAGS`.

mihe: Oh wait, I had completely glossed over the fact that this was just temporary until they merge that PR on their end.

Nevermind!

CMakePresets.json
```diff
@@ -254,11 +254,16 @@
...
       "hidden": true,
       "environment": {
+        "CFLAGS": "$penv{CFLAGS}",
         "CXXFLAGS": "$penv{CXXFLAGS} -msimd128 -msse4.2",
         "LDFLAGS": "$penv{LDFLAGS} -msimd128 -msse4.2 -Wno-limited-postlink-optimizations -sALLOW_MEMORY_GROWTH=1 -sALLOW_TABLE_GROWTH=1"
       },
       "cacheVariables": {
-        "CMAKE_TOOLCHAIN_FILE": "$penv{EMSDK}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake"
+        "CMAKE_TOOLCHAIN_FILE": "$penv{EMSDK}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake",
+        "EMSCRIPTEN_SYSTEM_PROCESSOR": "",
```
tetious, line 263: Note that I tried a bajillion different things to try to prevent needing this and spent far too long trying to figure out the proper way to check if a variable is uninitialized in Cmake. As far as I can tell, the behavior of the below idiom changed between v3.28.2 (on my Fedora 40 box) and v3.29 on the build machines. 

`if ("${EMSCRIPTEN_SYSTEM_PROCESSOR}" STREQUAL "")`

It blows up with an error (even when I disable --warn-uninitialized) and on my machine it works perfectly regardless of whether --warn-uninitialized is set or not.

I was unable to find any reference in the docs as to if this actually changed in 3.29, and I tried vendoring the toolchain and was completely unable to find any documentation as to how to detect an uninitialized variable in CMake. The most obvious: 

`if (NOT DEFINED EMSCRIPTEN_SYSTEM_PROCESSOR OR "${EMSCRIPTEN_SYSTEM_PROCESSOR}" STREQUAL "")`

Still explodes. So. This will work, I think. (at least it works in Act!)

Good enough for CI???? :worried:

mihe: Interesting. I just updated to CMake 3.29 from 3.28 and can confirm that `--warn-uninitialized` catches these cases now. It should rightfully do so as well, since the proper way to check for uninitialized variables is using `NOT DEFINED SOME_VARIABLE`, since `"${SOME_VARIABLE}" STREQUAL ""` would catch variables that are technically initialized as well, as you seem to have found out already.

> It blows up with an error in the CI CMake (even when I disable --warn-uninitialized)

I believe `--warn-uninitialized` is implied when passing `-Werror=dev`, so you would need to disable/remove that one as well. I get no warning when both of those are disabled.

The real problem here is of course that we can't tell CMake to ignore warnings in the toolchain file, and it obviously can't know whether we wrote it ourselves or not.

Honestly I'd be down to just disable both of these flags instead. Nobody ever enables them anyway, as evidenced by none of the Emscripten authors catching this. Even if we define these particular variables someone is going to use more uninitialized variables down the line and break the CI or whatever.

mihe: I just merged a new `strict` parameter to the `cmake-configure` action in #905, so you should be able to just do `strict: false` for web builds and you won't need these empty variables anymore.

tetious: Unfortunately this doesn't seem to work as expected. If I pass strict: false, it still errors.

I actually would prefer fixing this "correctly" though, and since we already have to hack around various problems with the toolchain file, I just vendored it.

`if (NOT DEFINED VAR)` was enough, without the extra or is empty. Turns out CMAKE does not short-circuit, so that's why my compound if wasn't working. :banana: 

Vendoring also lets me clean up the various hacky things I added all over the place, and feels overall much more clean.

I will see if I can get a PR through to fix this upstream, but in the meantime...

mihe: > If I pass strict: false, it still errors.

This doesn't make much sense to me. Plenty of projects are littered with use of uninitialized variables. I can't imagine CMake decided to suddenly make that a hard error without something enabling `--warn-uninitialized` and/or `-Werror=dev` somewhere somehow.

I am seeing mentions that `--warn-uninitialized` does in fact only check files that are at or under the project root (`CMakeLists.txt`), so I think part of the problem here is that the `EMSCRIPTEN_CACHE_FOLDER` is *in* the checkout folder as opposed to next to it (or above it) in the folder tree. Maybe there's another folder that could be used to house it, like `${{ runner.temp }}`? Although I don't know how that plays with the persistent cache.

> Vendoring also lets me clean up the various hacky things I added all over the place, and feels overall much more clean.

I don't love the idea of vendoring this, to be honest. I would have preferred the previous hacks, assuming that the uninitialized variable weirdness can be resolved in some way.

tetious: > Maybe there's another folder that could be used to house it

I can explore this possibility. I think I can set the cache folder to anything.

 > I don't love the idea of vendoring this, to be honest.

I don't either, though considering the staleness of the PR fixing shared library builds, I do wonder if the Cmake support is actually maintained within Emscripten.

I can switch back to the hackified version. Worst case, I got it to work by stubbing all the variables out to empty string, and that remains viable. (though icky...)

Your checkout folder idea sounds promising though, so I'll try that first.

Faless:
I've successfully tested this on single threaded builds from the latest main.

I've also also created this patch to add a separate profile for threaded builds, which can works with emscripten 3.1.61+git (see https://github.com/godotengine/godot/pull/93143 ).

The patch also assumes https://github.com/godotengine/godot/pull/93556 to allow the two binaries to co-exists and be exported according to the chosen variant.

[threads.diff.txt](https://github.com/user-attachments/files/15958759/threads.diff.txt) (NOTE: In the patch, I'm abusing the `cacheVariables` to set our own `EMSCRIPTEN_USE_PTHREADS`, which I later use to set the `-lpthread` flag and the files suffixes, not sure if there's a better solution)

tetious:
Awesome! Thanks very much, @Faless! 

Do you happen to know if 4.3 final's export templates will be built with Emscripten 3.1.62 (or +git)? OR if the output of 3.1.61+git works with the existing export templates (single threaded)? 

I will pull this patch into the PR, unless @mihe would prefer it split into a new PR?

I'm hopeful the export template issues can be ironed out before 4.3 so we can merge and release (at least with an experimental flag or something) near final 4.3-time.

mihe:
> I will pull this patch into the PR, unless @mihe would prefer it split into a new PR?

No, by all means, I'd prefer to see the web support finished as much as it can be in this PR.

We'll iron out whatever kinks there might be after it's been pulled in.

mihe:
Looks like you'll likely need to add `-Wno-experimental` among the `target_compile_options(...)` in order to get around `error: -sSIDE_MODULE + pthreads is experimental`.

CMakePresets.json
tetious: Alas, the checkout folder idea didn't pan out. The Emscripten action does not appear to work with anything outside the source root. 

I might be missing something, but I think the workaround Good Enough(TM) for now.

mihe: I just tried running the CI in regular GitHub Actions (not Act) without the CMake variables defined in and everything configures and builds just fine, with your latest commit, no errors no nothing.

tetious: Huh. That's weird... I started the whole journey because CI was going boom here on Github, and used Act to help me figure it out. Maybe you did something slightly differently?

Here is an example failure from the job history: https://github.com/godot-jolt/godot-jolt/actions/runs/9571585844/job/26389026062#step:4:114

Oh wait, do you mean AFTER the strict mode change? I don't think I ever tried that outside Act... Hmmmmm...

mihe: > Oh wait, do you mean AFTER the strict mode change?

Well, yes, technically, but I left strict mode enabled. I can see the respective CMake flags being passed to it.

Here's the run, assuming you have permission to see it: https://github.com/mihe/godot-jolt/actions/runs/9653686082

.github/workflows/build-web.yml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+  setup_emscripten:
+    name: Pre-warm cache for emscripten
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up emsdk
+        uses: mymindstorm/setup-emsdk@v14
+        with:
+          version: ${{ env.EMSCRIPTEN_VERSION }}
+          actions-cache-folder: ${{ env.EMSCRIPTEN_CACHE_FOLDER }}
```
mihe, line 39: I'm not sure this does what you want it to do. There's no dependency set up between this job and the `build` job, so all the build jobs can/will run before the pre-warming has run.

I suspect you'll need `needs: setup_emscripten` in the `build` job.

tetious: Indeed. Will fix. Thanks!

CMakePresets.json
tetious: I think it is exactly what you suspected, and since it wasn't cached, the folder is outside the build root and so the check doesn't happen.

Your toolchain file:
` CMAKE_TOOLCHAIN_FILE="/home/runner/work/_temp/ee5c6111-c45f-4c92-9497-edd8179428ec/emsdk-main/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake"`

This PRs, with working cache (due to it running more than once, I guess?):
`CMAKE_TOOLCHAIN_FILE="/home/runner/work/godot-jolt/godot-jolt/emsdk-cache/emsdk-main/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake"`

I am also suspecting now that the cache directory error I was seeing when trying to use the tmp folder was either an Act thing or my not setting the `needs:`

I will try again. Thanks for all your help on this!!

tetious: :( 
`const fullCachePath = path.join(process.env.GITHUB_WORKSPACE, emArgs.actionsCacheFolder);`

mihe: Honestly, judging by how quickly these jobs completed even without the cache being used, I would say just ditch `actions-cache-folder` thing altogether. Setting up Emscripten is never going to be a bottleneck for CI. That would then presumably make it use this temporary directory instead and fix all this.

tetious: > What's a few more minutes when you have an hour-long CI. :)

Haha! Good point! :laughing: 

Done!

mcjohnalds:
I tested the joins.tscn example scene on Windows and an M1 Mac on the following browsers:

- Edge
- Safari
- Firefox
- Chrome

And it worked flawlessly! Really impressed. Amazing thanks to tetious and the godot-jolt contributors. I will be using jolt in game jams from now on.

mihe:
I've been meaning to try out this PR for myself, as I was planning on moving some things around, but time has been scarce.

What I'd like to see changed is (and I'm happy to do this myself when I have time):

1. The CMake presets aren't really set up in the way they probably/unfortunately need to be. Since this is effectively cross-compilation, the presets ought to be more like the Android presets, which (again, unfortunately) are in the format of `[host]-android-[arch]` and `[host]-android-[arch]-[config]` for the configuration and build presets respectively. Without this you're not able to build for web on anything but Linux, since they inherit from `linux-base`, which enforces that it's only available on that platform.
2. This is somewhat arbitrary, but threads on/off probably shouldn't be part of the presets. I try to avoid adding to the combinatorial explosion as much as I can, and I think it's best if threading is made into a CMake option instead, much like `GDJ_DOUBLE_PRECISION`. It could (should?) perhaps even be a general flag that forces usage of `JPH::JobSystemSingleThreaded` instead of `JoltJobSystem`, but that'll probably require some refactoring of the code as well.
3. The non-threaded binaries should use the [newly introduced](https://github.com/godotengine/godot/pull/93563) `nothreads` feature tag.
4. The compilation flags that are currently being added through `CMakePresets.json` all need to be moved into `CMakeLists.txt`. It should (in theory at least) be possible to use some other CMake generator (without the presets) and just pass the toolchain file and have things work. If some of those flags need to propagate to Jolt or some other dependency then they need to be forwarded through their respective `GodotJoltExternal*.cmake` file or be made into an upstream feature request.
5. The `-msse4.2` flag still bothers me, and I can't imagine it's actually needed.
6. I can't help but feel like we should just pass `-sINITIAL_MEMORY` for both threaded and non-threaded, instead of having different behavior for different builds.
7. Packaging still needs to be done I guess.

I'm also curious whether the amount of threads reported by Godot is correct for `threads=yes` builds, considering the comment you (@tetious) made [here](https://github.com/godot-jolt/godot-jolt/issues/477#issuecomment-2174581777) about it returning 2 for you with `threads=no`.

Faless:
> 6\. I can't help but feel like we should just pass `-sINITIAL_MEMORY` for both threaded and non-threaded, instead of having different behavior for different builds. There's also the question of what the right value is for this argument.

About this, we shouldn't need to specify the `INITIAL_MEMORY` when building extensions (`SIDE_MODULE`s),

tetious:
@mihe When you get a sec, would you mind updating the jolt fork, please? The fix is in for excluding `-mfpmath=sse` for Emscripten builds.

Thanks!

jrouwe:
> The -msse4.2 flag still bothers me, and I can't imagine it's actually needed.

@mihe It actually is, if you compile with the emscripten flags `-msimd128 -msse4.2` then it will convert SSE intrinsics to WASM SIMD instructions. Without it you fall back to plain C++ math because I didn't implement the WASM SIMD intrinsics.

mihe:
> @mihe It actually is, if you compile with the emscripten flags `-msimd128 -msse4.2` then it will convert SSE intrinsics to WASM SIMD instructions. Without it you fall back to plain C++ math because I didn't implement the WASM SIMD intrinsics.

Sorry, I ended up replying in [that Jolt issue](https://github.com/jrouwe/JoltPhysics/issues/1173#issuecomment-2226503019) before I saw you reply here.

My poorly phrased comment was meant as a continuation of [this review comment](https://github.com/godot-jolt/godot-jolt/pull/902#discussion_r1644899262). My concern was more about why it's SSE 4.2 specifically and not SSE2, and why Jolt requires passing it rather than adding it itself, like with the other platforms.

I wasn't aware WebAssembly had its own set of SIMD instructions though. Is that somehow influencing the reasoning behind passing the x86 flags manually?

jrouwe:
> I somehow missed that WebAssembly had its own set of SIMD instructions though. Is that somehow influencing the reasoning behind passing the x86 flags manually?

I'll reply in the other ticket.

tetious:
Just want to make a quick update to note that this PR is not abandoned. My project (using this) has a milestone approaching that has filled all my time for the last month. The good news is that I've been dogfooding the web support heavily and it is working GREAT, with the exception of Trimesh colliders on Chromebooks. I still need to figure out why any trimesh causes an insta memory access violation on that platform. 

Anyway, after 8/27 I should have some time and can hopefully wrap this up.

mcjohnalds:
I can confirm this PR works great with the standard Godot v4.3.stable web export templates.

I tested this with  Chrome, Firefox, Edge, and Safari on Windows and Mac.

I'm shook.

tetious:
I have now updated the zip in the PR description with a new build, based on current master. (hash 5497271) This new build is very lightly tested, but I'll be testing it more soon.

@mihe I'm going to need some help/advice restructuring the build so it works as you requested. The two current problems I don't know how to solve in CMAKE:

- Anything built with SSE2 needs `-msimd128 -msse4.2` due to how Emscripten implements SIMD. Jolt handles this internally now, and I *think* I figured out how to make this work when building the extension itself, but I'm not sure how to proceed with godot-cpp. I think it will need a similar tweak to [this from jolt](https://github.com/jrouwe/JoltPhysics/blob/7ca910cb4d21627960f4fcc0c22615343e7327c3/Jolt/Jolt.cmake#L642C39-L642C57).
- When building with thread support, `-pthread` is not enough. A combination of additional options need to be set when compiling and linking (shared memory, etc) that are wrapped up by emscripten in `-sUSE_PTHREADS=1` (godot uses it [here](https://github.com/godotengine/godot/blob/61598c5c88d95b96811d386cb20d714c35f4c6d7/platform/web/detect.py#L225)) This needs to be passed globally, and I can't figure out a way to accomplish this in CMakeLists.txt. I tried many, many things from Googling. :disappointed: 

I'm hoping you can suggest a path as I'm struggling to figure out some of the stuff you're doing with CMake. I've rarely used it, and I've exhausted my Google-fu. Something about how you're building the external libraries stomps on all attempts to tweak the compile/link flags that I've tried, except the env vars (or using cmakepresets)

Thanks for your help!

mihe:
> Something about how you're building the external libraries stomps on all attempts to tweak the compile/link flags that I've tried, except the env vars (or using cmakepresets)

The way I deal with the external libraries is somewhat unusual, to say the least. The original motivation behind it, as opposed to CMake's `FetchContent`, was to be able to have a different set of `CMAKE_CONFIGURATION_TYPES` from what the external libraries have. That simple requirement has some pretty severe ramifications, "forcing" me to use its `ExternalModule` stuff directly instead, which essentially kicks off an entirely separate CMake environment when building the external libraries. It does also have the added "benefit" (depending on how you look at it) of building every dependency as top-level, which is something a lot of CMake-based libraries out there tend to assume is the case, and typically break when it isn't.

Anyway, you're right in that this would require some changes to my fork of godot-cpp to append the SIMD-related flags, similar to what Jolt is doing with the `USE_WASM_SIMD` option now. You can technically do the environment stuff in a more isolated fashion by doing what I do with mimalloc already (see [here](https://github.com/godot-jolt/godot-jolt/blob/54972714d9699a9e241e47aaa61a80561d56afaa/cmake/GodotJoltExternalMimalloc.cmake#L16-L26) and [here](https://github.com/godot-jolt/godot-jolt/blob/54972714d9699a9e241e47aaa61a80561d56afaa/cmake/GodotJoltExternalMimalloc.cmake#L37-L38)), but that seems contrived given that we have full control over the godot-cpp fork. You can just leave that bit (including the threading flags) for now and I can fix it instead, along with the godot-cpp changes.

tetious:
(apologies for the force-push; only the last commit is actually different from the previous state of the branch)

mihe:
First of all, apologies for not tackling some of the issues here that I said I would take a look at. I realize that this PR likely has just needed a few days of me sitting down and taking it across the finish line, but that never ended up happening for a variety of reasons.

You might perhaps have seen this already, @tetious, but this extension has now been merged into Godot in the form of an engine module instead. As a result, and as mentioned in #477, the web support there largely comes "for free" due to being part of Godot's build and release process. I've only tested it very briefly myself, but both threaded and non-threaded builds seem to work, although I don't have access to ChromeOS to test the issue you ran into there.

I also haven't looked into whether Jolt's WASM SIMD support is properly enabled in that new engine module. As mentioned here previously, Godot does seem to pass `-msimd128` as part of building for web, but I believe it only passes `-msse2` as opposed to `-msse4.2`, so there might perhaps be some low-hanging fruit there in terms of performance, I'm not sure.

Also note that I still make use of Jolt's `JPH::JobSystemWithBarrier` even for non-threaded builds, albeit with a reported `GetMaxConcurrency()` of `1`. Again, things seemed to work during my very brief testing, but there might perhaps be some concerns there. If nothing else, I imagine that Jolt's `JPH::JobSystemSingleThreaded` would be a more sensical for those builds, and perhaps perform better.

Anyway, this extension will be considered to be in maintenance mode from now on (see [README](https://github.com/godot-jolt/godot-jolt?tab=readme-ov-file)) where only bug fixes will be considered for merging. As such, I'll go ahead and close this pull request.

Thank you for the considerable effort you put into this @tetious, and I hope this work was of some use to you at least.

PR #941, opened 2024-08-17
Added support for `WorldBoundaryShape3D`

### LINKED ISSUE #104: WorldBoundaryShape3D
## Description

This issue tracks the progress of implementing the functionality of Godot's `WorldBoundaryShape3D` node type.

## Status

- **Not supported** 👈
- ~~Not started~~
- ~~Partial (not functional)~~
- ~~Partial (functional)~~
- ~~Done (as much as it can be)~~
- ~~Done (completely)~~

## Links

Documentation: [`WorldBoundaryShape3D`](https://docs.godotengine.org/en/latest/classes/class_worldboundaryshape3d.html)
Inherits: [`Shape3D`](https://docs.godotengine.org/en/latest/classes/class_shape3d.html)

mihe:
Fixes #104.

This implements `WorldBoundaryShape3D` using the newly introduced `JPH::PlaneShape` as the underlying Jolt shape.

This also brings with it a project setting for setting the effective size of this plane, which by default is set to 2000 meters.

### DIFF
--- a/src/objects/jolt_shaped_object_impl_3d.cpp
+++ b/src/objects/jolt_shaped_object_impl_3d.cpp
@@ -366,7 +366,7 @@ void JoltShapedObjectImpl3D::post_step(float p_step, JPH::Body& p_jolt_body) {
 bool JoltShapedObjectImpl3D::_is_big() const {
 	// HACK(mihe): This number is completely arbitrary, and mostly just needs to capture any
 	// `WorldBoundaryShape3D`. There could be a better sweet spot to be found here.
-	return get_aabb().get_longest_axis_size() >= 900.0f;
+	return get_aabb().get_longest_axis_size() >= 1000.0f;
 }
 
 JPH::ShapeRefC JoltShapedObjectImpl3D::_try_build_single_shape() {
--- a/src/servers/jolt_project_settings.cpp
+++ b/src/servers/jolt_project_settings.cpp
@@ -38,6 +38,7 @@ constexpr char BOUNCE_VELOCITY_THRESHOLD[] = "physics/jolt_3d/solver/bounce_velo
 constexpr char CONTACT_DISTANCE[] = "physics/jolt_3d/solver/contact_speculative_distance";
 constexpr char CONTACT_PENETRATION[] = "physics/jolt_3d/solver/contact_allowed_penetration";
 
+constexpr char WORLD_BOUNDARY_SIZE[] = "physics/jolt_3d/limits/world_boundary_shape_size";
 constexpr char MAX_LINEAR_VELOCITY[] = "physics/jolt_3d/limits/max_linear_velocity";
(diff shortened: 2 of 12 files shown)


src/objects/jolt_shaped_object_impl_3d.cpp
```diff
@@ -345,6 +345,18 @@ void JoltShapedObjectImpl3D::post_step(float p_step, JPH::Body& p_jolt_body) {
...
+	for (const JoltShapeInstance3D& shape : shapes) {
+		if (shape.is_enabled() &&
+			shape.get_shape()->get_type() == PhysicsServer3D::SHAPE_WORLD_BOUNDARY)
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
```
mihe, line 358: @jrouwe Would you say it makes sense to generalize this method into calculating the AABB for the body as a whole (while caching the ones from the complex shapes) and using some arbitrary size as the cut-off to determine whether a body is "huge"?

That way you wouldn't only have `PlaneShape` bodies in this new broadphase layer, but also whatever other "huge" static body that the user decided to add to their world, which might perhaps speed things up a little bit in the regular `BODY_STATIC` broadphase layer?

I'm thinking there would only ever really be a handful of bodies in this new `BODY_STATIC_SLOW` layer either way, say maybe 6x `PlaneShape` and some huge `MeshShape` or two at most.

jrouwe: Yes, I think that makes sense.

src/servers/jolt_project_settings.cpp
```diff
@@ -170,6 +171,7 @@ void JoltProjectSettings::register_settings() {
 	register_setting_ranged(PAIR_CACHE_DISTANCE, 0.001f, U"0,0.01,0.00001,or_greater,suffix:m");
 	register_setting_ranged(PAIR_CACHE_ANGLE, Math::deg_to_rad(2.0f), U"0,180,0.01,radians");
 
+	register_setting_ranged(WORLD_BOUNDARY_SIZE, 2000.0f, U"1,2000,or_greater,suffix:m");
```
mihe, line 174: @jrouwe Seeing as how we now have a dedicated broad phase layer for these big objects anyway, is there any major downside to just cranking this up to something silly, like what Godot/Bullet uses? That way we could presumably just omit this setting altogether.

Obviously you'll still run into "natural" precision problems once you move far enough away from (0, 0, 0), but would a half-extent of (for example) `1e+8` force any additional problems, even if you're sticking reasonably close to (0, 0, 0)?

jrouwe: The GetSupportingFace, GetTrianglesStart and indeed the Draw functions become quite inaccurate if you bump the value up that high. You're not calling the first 2 functions right now, but if you ever decide to implement CharacterVirtual then it will call into the GetSupportingFace function. I would be inclined to keep the setting for now.

Issue #351, opened 2023-05-18
Add a node for Jolt's `DistanceConstraint`


mihe:
Jolt's `JPH::DistanceConstraint` is described as:

> Will attach two bodies with a stick (removing 1 degree of freedom).

I figured since there's no real equivalent to this fairly basic constraint in Godot, without resorting to the more expensive 6DOF joint, it might be good to add this as `JoltDistanceJoint3D` or something.

This joint also allows for springy behavior using its frequency and damping properties, which can be useful for things like picking.

ODtian:
This feature is very common in other engines but missing in godot, is this still planning in godot-jolt?

mihe:
It's highly likely that this extension will be moved over to be an official Godot module in a not too distant future, so I will be focusing solely on stuff that helps close the gap with Godot Physics going forward, meaning this will likely not be worked on.

Your best bet is probably to create a Godot proposal instead.

mihe:
As mentioned here already, this is very unlikely to be worked on at this point, so I'll go ahead and close this.

PR #976, opened 2024-09-05
Adding distance constraint

Ewall198:
#issue 351

Adding `jolt`'s `DistanceConstraint` to `godot-jolt` as `JoltDistanceConstraint3D`.
This includes:
- A new icon
- Gizmo display
- Added an example of usage of `JoltDistanceConstraint3D` to the `joints.tscn` packed scene. (As well as how to draw a distance constraint in game `DistanceConstraintDisplay3D`)

Notes on implementation:
This joint mimics the existing joints in class structure in order to make it easier to wrap the Godot distance constraint if/when it is implemented.
If NodeA or NodeB is not set, the Constraint will attach the body to a world point. If the `JoltDistanceConstraint3D` moves while running, the world point which the body is attached to will not move. (Chose this implementation for the sake of performance).

Demo - See the two cubes in the upper right.
distance_constraint_demo

Side note:
When using `JoltDebugGeometry3D` with `JoltDistanceConstraint3D`, it will try to use `DrawText3D` which isn't implemented. In order to handle this, I Added a flag `draw_text` which can be toggled by the user (and defaults to false). This way the user won't get spammed with "Not implemented" errors. I started an implementation of `DrawText3D` [here](https://github.com/Ewall198/godot-jolt/tree/segment-display), but stopped because it ended up being a bit more complicated than I expected and this PR was big enough.


mihe:
I greatly appreciate the work you've put into this, and it does seem very well done, but I must admit I'm a little bit apprehensive about merging this.

The reason for closing #351 as "not planned" wasn't so much that I don't have time to do this (although that's certainly true to some degree) but more that since a few months back now I've actively avoided adding features to this extension that aren't present in Godot Physics[^1].

The reason for this being that this extension will likely be migrated over to Godot as an (experimental) engine module fairly soon, likely within the next few weeks (after which this extension will be put into maintenance mode and archived some time later), which will will lead to some tough questions about how these two engines will co-exist, which would only be made worse by this extension being a superset of Godot Physics.

Once Jolt has been deemed to sufficiently supersede Godot Physics 3D (likely months down the road, perhaps longer) and no more (or at least very few/insignificant) issues are being reported about discrepancies between the two engines, it will likely be made the default 3D physics engine in Godot and Godot Physics 3D will likely be deprecated, at which point it seems much more appropriate to start talking about adding features to Godot that are Jolt-specific (like buoyancy, all its additional constraints, better vehicle support, character controller, etc. etc.)

Now, to be fair, this constraint could (much like the slider/hinge/cone-twist joints) be expressed through a 6DOF joint already (see my [picking](https://github.com/godot-jolt/godot-jolt/blob/1bed4e48e60650241cd0dbba66ecb1ac3d778eb1/examples/scenes/common/scripts/free_look_camera.gd#L42-L56)) so it would technically be possible to implement for Godot Physics by just using a 6DOF joint behind the scenes as a placeholder implementation, but it'll still be somewhat incomplete since it won't have support for springs, which I would argue is a core part of this particular constraint. Also, in the (highly theoretical/unlikely) situation where Jolt isn't in fact picked to be the next 3D physics engine in Godot then Godot Physics finds itself in a somewhat weird spot.

Anyway, I'm curious to hear how you feel about this. The options as I see them are either:

1. Close this PR for now, and you can instead make a PR to add this to Godot itself, once my upcoming PR for integrating Jolt has been merged, and your PR for Godot can then be merged whenever deemed appropriate.
2. Merge this PR and just not bring this particular joint along in the migration.
3. Merge this PR and provide a (incomplete) 6DOF-backed implementation for Godot Physics as part of the Godot migration.
4. Let this PR sit open/unmerged for whomever wants to build your fork to make use of it.

I would probably prefer option 1. How would you feel about that?

[^1]: The enhanced edge removal would be the exception I guess, but that's just a single project setting that slightly changes a behavior.

Ewall198:
Thanks for the response, this makes a lot of sense! 1 sounds great to me (with the added bonus of getting to say I contributed to `godotengine/godot` 😄).

This is a little nit-picky, but it's worth noting that a 6DOF joint cannot quite replicate the behavior of a distance constraint.
6DOF - If you set linear limits of x, y, & z to [-1, 1] then the bounds of the constraint act like a **cube**.
6DOF - If you set the linear limits of x & z to 0 and linear limits of y to [-1, 1] the bounds of the constraint act like a **cone** pointing down.
DistanceConstraint - the bounds of the constraint are shaped like a **sphere**.

Thanks for taking a look! I'll try to keep up to date on updates for Godot and Godot-Jolt, but please let me know if there is an issue I can track, or just let me know when is a good time to resubmit this PR to `godotengine/godot`.

mihe:
> DistanceConstraint - the bounds of the constraint are shaped like a **sphere**.

That's a good point!

Even when disregarding the limit shape I'm realizing now that the fundamental aspect of the distance constrained body being able to rotate freely around itself couldn't actually be satisfied with a single 6DOF joint, so that was just a bad take on my part in general.

Anyway, thank you for your understanding.

> please let me know if there is an issue I can track, or just let me know when is a good time to resubmit this PR to `godotengine/godot`.

You can keep an eye on [this proposal](https://www.github.com/godotengine/godot-proposals/issues/7308), which will likely be closed as part of the eventual PR that migrates this extension to Godot.

i-draw-boxes:
Sorry for asking in this old PR, but I think it fits here best. 

Do you know of any way how to currently produce the expected behavior - DistanceContraint with min/max movement that removes only a single DoF - with the current tools that godot-jolt gives us? As was mentioned before, the available space for the joined object would be defined by a sphere with radius maxRadius and a sphere with the radius minRadius.


You mentioned in your response that it can't be done with a single 6DoF joint, but I don't really see how having multiples would help...

Ewall198:
@i-draw-boxes sorry, I don't know of any way to achieve the functionality of the distance constraint with the current `godot-jolt` tools. Your best bet is probably to use 6DOF joint and fiddle with the settings, since it will get you pretty close and might look almost correct to the end user. You could try using multiple 6DOF joints to sort of round out the corners of the bounding cube.

Your other options are: 
- use this branch directly.
- Implement your own elastic connection. Elastic connections are super easy. Just take the distance between two points (of two objects) and whenever that distance exceeds your max distance by `delta_x` you apply a force at those points towards each other `F = k * delta_x` where k is the elasticity.

i-draw-boxes:
@Ewall198 

I'm using a spring as you discribed and it works well enough for the prototype. It also makes it easy to have minimum and maximum distance limits and plays well enough with the other joints that are constraining one of the two bodies. 

Thank you for helping me out with this one!

PR #987, opened 2024-09-24
Added ARM64 port that compiles and installs

### LINKED ISSUE #981: Add Linux ARM builds
Or so I heard... Raspberry needs your help!

bkeys:
It was as simple as removing the MSSE when it's not an X64 CPU. This should also fix #981


mihe:
Closing this, as per my comment [here](https://github.com/godot-jolt/godot-jolt/issues/981#issuecomment-2556969893).

PR #997, opened 2024-10-11
Fix reported joint force/torque not being correct when sprung

### LINKED ISSUE #993: get_applied_force() of JoltGeneric6DOFJoint3D not working as expected
Hello everyone,

Thanks for providing this valuable Godot addon. I'm currently working on a small bridge building game, similar to _Bridge It_ or _Poly Bridge_. For the physics I would like to use Jolt, after some testing this seems to be the best fit for my needs. I'm using JoltGeneric6DOFJoint3D to connect the parts of the bridge.
Now I'm struggling with implementing breakable joints and calculating the stress a single element of the bridge is experiencing (for visualisation). My current approach is to use the get_applied_force method of the joints. But it seems to me that this is not working, at least not as I would expect.

I have prepared a demo to illustrate the problem. The red boxes are the anchors created using StaticBody3D. The long yellow boxes are the beams created using RigidBody3D. The adjacent anchors and beams are connected using three JoltGeneric6DOFJoint3D, without any angular DOF constraints, but with either "Linear Limit" **or** "Linear Spring" active on all axes.
I have set Can Sleep to false for the physics nodes.
A Label3D is used to display the value from the get_applied_force method of each joint.

demo
[demo-project-to-illustrate-the-issue.zip](https://github.com/user-attachments/files/17249494/physics-issue.zip)

For my game I prefer to use the "Linear Spring" variant, but this always gives me 0.0 for the applied force. Is this expected? Are the high values expected in the "Linear Limit" variant and btw what is the unit of this number?

Thanks in advance!

"Linear Limit"  variant   |  "Linear Spring" variant 
:-------------------------:|:-------------------------:
grafik |  grafik

jrouwe:
* Hinge and slider joints weren't including the lambda of the limits
* Including motor lambda for all joints

Fixes #993

### DIFF
--- a/src/joints/jolt_cone_twist_joint_impl_3d.cpp
+++ b/src/joints/jolt_cone_twist_joint_impl_3d.cpp
@@ -212,17 +212,18 @@ float JoltConeTwistJointImpl3D::get_applied_torque() const {
 	const float last_step = space->get_last_step();
 	QUIET_FAIL_COND_D(last_step == 0.0f);
 
-	const Vector3 rotation_lambda = Vector3(
+	const Vector3 swing_twist_lambda = Vector3(
 		constraint->GetTotalLambdaTwist(),
 		constraint->GetTotalLambdaSwingY(),
 		constraint->GetTotalLambdaSwingZ()
 	);
 
-	// The motor lambda is in a different space than the swing twist lambda, and since the two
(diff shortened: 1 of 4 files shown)


mihe:
Thank you for looking into this. Much appreciated.

PR #1017, opened 2024-11-17
Fixed issues related to enhanced internal edge removal

### LINKED ISSUE #1014: `intersect_shape` sometimes omits expected contacts
Is this a known feature/setting/issue where the max_result parameter of `intersect_shape()` is ignored if it's set manually to higher than the default 32?
I'm encountering this in a custom double-precision v4.3 build, GodotPhysics handles it correctly..

I thought I asked before creating a MRP, but will be happy to do so

mihe:
Fixes #1014.

This pull request mainly does two things:

1. It fixes the incorrect use of Jolt's enhanced internal edge removal in the context of shape queries hitting multiple bodies, where `JPH::InternalEdgeRemovingCollector` apparently has to be flushed inbetween every body. This is now done using the new `CollideShapeWithInternalEdgeRemoval` method.
2. It splits the current "Use Enhanced Internal Edge Detection" setting into three settings, in order to be able to distinctly control its use for `RigidBody3D`, kinematic queries (`move_and_slide`, etc.) and shape queries (`intersect_shape`, etc.), since its effects on shape queries can be undesirable when not using them for kinematic movement purposes.
3. It changes the default value of this setting for queries to be off, in order to always return all hits by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,31 @@ Breaking changes are denoted with ⚠️.
 
 ## [Unreleased]
 
+### Changed
+
+- ⚠️ Changed the "Use Enhanced Internal Edge Detection" project setting under the "Collisions"
+  category to only apply to the collision detection that happens for `RigidBody3D`. There are now
+  instead identically named settings under the "Kinematics" and "Queries" categories for controlling
+  those contexts. See below for more information.
+
+### Added
+
+- Added a new "Use Enhanced Internal Edge Detection" project setting under the "Kinematics" category
+  for controlling the use of Jolt's enhanced internal edge detection in the context of kinematic
+  movement methods, meaning the `body_test_motion` method of `PhysicsServer3D`, which is also what
+  powers `test_move`, `move_and_collide` and `move_and_slide`.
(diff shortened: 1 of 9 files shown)


jrouwe:
> Can you really make the assumption with Jolt (when not using InternalEdgeRemovingCollector) that you'll register a hit against every "leaf" shape of a body when doing a CollideShape query (assuming of course that the query shape overlaps with them all)?

That really depends on your collector and what you do with `UpdateEarlyOutFraction`. If you use the `AllHitCollisionCollector` then you would get one or more hits for every leaf shape. If you use `ClosestHitCollisionCollector` then it tries to be smart. For a compound shape it visits the closest shapes first and if a hit is found then it will no longer process the other leaf shapes when the hit cannot be closer (due to the bounding box being further away than `GetEarlyOutFraction`).

mihe:
Right, yeah, I meant for something like a `AllHitCollisionCollector`.

In that case this change seems reasonable, but I'll await your helper function, since it would slim this down a bit.

jrouwe:
I think jrouwe/JoltPhysics#1346 should give you the interface you need.

mihe:
Yes, that looks great, thank you!

PR #1028, opened 2024-12-14
Implemented waking up body's neighbors

TacoGuyAT:
Currently, when PhysicsBody3D is moved/freed, its neighbors remain floating in the air. This PR adds an utility method to manually wake up neighbors using broad phase query (https://github.com/jrouwe/JoltPhysics/issues/267#issuecomment-1274151892).

### DIFF
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -257,9 +257,7 @@ void JoltBodyImpl3D::set_is_sleeping(bool p_enabled) {
 }
 
 void JoltBodyImpl3D::wake_up_neighbors() {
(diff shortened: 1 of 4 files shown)


mihe:
Thank you for the contribution. I can certainly see how this could be useful.

However, this extension has just recently been integrated into Godot in the form of an engine module, and as such this extension will be considered to be in maintenance mode going forward. This means that I will only accept pull requests for bug fixes.

(I've updated `README.md` with this information now as well.)

If you want to see this be added to Godot itself I would recommend making [a proposal](https://github.com/godotengine/godot-proposals) and move your pull request to Godot's repository instead. Note that this would/should likely involve implementing this for Godot Physics as well, which might be non-trivial.

PR #1029, opened 2024-12-15
Added updating of damping when switching damping mode

opsocket:
Hi there 👋

Here's a fix for an issue where [linear_damp](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-linear-damp) and [angular_damp](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-angular-damp) are not updated when either [linear_damp_mode](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-linear-damp-mode) or [angular_damp_mode](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-angular-damp-mode) is changed at runtime.

Previously, [linear_damp](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-linear-damp) and [angular_damp](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#class-rigidbody3d-property-angular-damp) were computed once using their initial [damping mode](https://docs.godotengine.org/en/latest/classes/class_rigidbody3d.html#enum-rigidbody3d-dampmode) and not updated whenever a mode was changed.

It's likely that we want to split that update operation for each damping type.

---

I've made a short demo, Cube A and B are the same but they change their linear damping mode when they collide with a floor surface, the red one changes the mode to REPLACE while the blue one changes the mode to COMBINE.
The initial mode is COMBINE for both and they have the same initial velocity in opposite directions. 
The default linear damping value is `.5`.

https://github.com/user-attachments/assets/c937b559-04fa-489d-a275-21bf57dde1bc

### DIFF
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -186,9 +186,11 @@ void JoltBodyImpl3D::set_param(PhysicsServer3D::BodyParameter p_param, const Var
 		} break;
 		case PhysicsServer3D::BODY_PARAM_LINEAR_DAMP_MODE: {
 			set_linear_damp_mode((DampMode)(int32_t)p_value);
+			_update_damp();
 		} break;
 		case PhysicsServer3D::BODY_PARAM_ANGULAR_DAMP_MODE: {
 			set_angular_damp_mode((DampMode)(int32_t)p_value);
+			_update_damp();
 		} break;
 		case PhysicsServer3D::BODY_PARAM_LINEAR_DAMP: {
 			set_linear_damp(p_value);
--- a/src/objects/jolt_body_impl_3d.cpp
+++ b/src/objects/jolt_body_impl_3d.cpp
@@ -1042,6 +1042,26 @@ void JoltBodyImpl3D::set_angular_damp(float p_damp) {
 	_update_damp();
 }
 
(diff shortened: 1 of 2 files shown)


mihe:
Hello hello!

I'm not sure if you're the same person who reported this bug to me on Mastodon a few days ago, but thank you for taking the time to fix it either way. I ended up not having as much time as I thought to actually push the fix these last few days.

> It's likely that we want to split that update operation for each damping type.

It's fine as it is. Updating the damping is not a particularly expensive operation.

> I've read your reply [here](https://github.com/godot-jolt/godot-jolt/pull/976#issuecomment-2333622990) and opened another [PR](https://github.com/godotengine/godot/pull/100473) on the engine but it may be a bit overkill because we now need to recompile the whole engine instead of the extension only

That reply only concerns "enhancements", meaning stuff that isn't just a bug fix, but actually adds to the feature set or the interface in some way. This PR would be a bug fix, and should indeed be fixed in this extension as well, as I plan to continue supporting the extension (in "maintenance mode") through at least Godot 4.4.

Note that the extension will be unsupported (and later archived) once the module integration reaches some kind of feature parity with the extension.

PR #1031, opened 2024-12-20
Added update of overlapping rigid bodies damping when `Area3D::*_damp_space_override` is changed at runtime

opsocket:
I already mentioned the issue shortly but I thought I'd sort it out before bothering you.

Briefly, it's the same issue but for areas. If the mode is changed at runtime then an area has to tell overlapping bodies to recompute total damping, as the space override mode might change the value applied to a body.

In fact, it could also be an issue for [gravity_space_override](https://docs.godotengine.org/en/stable/classes/class_area3d.html#class-area3d-property-gravity-space-override).

### DIFF
--- a/src/objects/jolt_area_impl_3d.cpp
+++ b/src/objects/jolt_area_impl_3d.cpp
@@ -3,6 +3,7 @@
 #include "objects/jolt_body_impl_3d.hpp"
 #include "objects/jolt_group_filter.hpp"
 #include "objects/jolt_soft_body_impl_3d.hpp"
+#include "servers/jolt_physics_server_3d.hpp"
 #include "servers/jolt_project_settings.hpp"
 #include "spaces/jolt_broad_phase_layer.hpp"
 #include "spaces/jolt_space_3d.hpp"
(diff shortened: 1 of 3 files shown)


mihe:
> Briefly, it's the same issue but for areas. If the mode is changed at runtime then an area has to tell overlapping bodies to recompute total damping, as the space override mode might change the value applied to a body.

This is indeed the same issue, and the fix looks about right, although I'll probably make some minor changes to it before merging.

> In fact, it could also be an issue for [gravity_space_override](https://docs.godotengine.org/en/stable/classes/class_area3d.html#class-area3d-property-gravity-space-override).

Yes. Frankly, the same applies to the [`linear_damp`](https://docs.godotengine.org/en/stable/classes/class_area3d.html#class-area3d-property-linear-damp), [`angular_damp`](https://docs.godotengine.org/en/stable/classes/class_area3d.html#class-area3d-property-angular-damp) and [`priority`](https://docs.godotengine.org/en/stable/classes/class_area3d.html#class-area3d-property-priority) properties as well.

I'm not sure why I didn't do this to begin with, but considering that this seems to be a bug/oversight in Godot Physics as well I'm guessing I might have decided to just emulate whatever they were doing there.

I'll throw in fixes for the above mentioned properties in this pull request as well.

mihe:
Apologies for letting this one sit for so long.

I'm about to put out a new release for the people who still need the joint changes in this extension, so I sat down and looked at getting this merged, along with the changes I mentioned above, but the more I looked at it the more I realized that all of this stems from the fact that I diverged from Godot Physics in how the calculation of total damping is done, where Godot Physics calculates it on every tick, thereby avoiding all of these issues (including #1029).

I decided to change to use that approach instead, since I'm a bit pressed for time, and I find it easier to reason about.

That means this PR is superseded, so I'll go ahead and close it.

